PR #280, opened 2021-11-02
ORDER BY parser and analyzer

### LINKED ISSUE #272: Add parser for ORDER BY
Have the following work:

```
user@user-komp:~/sources/opendiffix/reference$ dotnet run --project src/OpenDiffix.CLI/ -- -f data/data.sqlite --aid-columns customers.id -q "SELECT age, count(*) FROM customers GROUP BY age ORDER BY age"
ERROR: Parse error: Error in Ln: 1 Col: 50
SELECT age, count(*) FROM customers GROUP BY age ORDER BY age
                                                 ^
Expecting: end of input, infix operator, postfix operator, ',', '.', ';',
'HAVING' (case-insensitive) or 'LIMIT' (case-insensitive)
```

Related follow-up issue: #256

pdobacz:
Closes #272. Adds parser and analyzer support for ORDER BY clauses.

handling `ORDER BY` by aliases to be tackled separately, for now as workaround use the target expression index as in

```
SELECT age, city, count(DISTINCT ID) as c FROM customers GROUP BY age, city ORDER BY 3
```

instead of

```
SELECT age, city, count(DISTINCT ID) as c FROM customers GROUP BY age, city ORDER BY c
```

I'm still scratching my head, if there are any anonymity-related concerns with ORDER BY, or things will "just work".

### DIFF
--- a/src/OpenDiffix.Core/Analyzer.fs
+++ b/src/OpenDiffix.Core/Analyzer.fs
@@ -142,6 +142,39 @@ let private mapOrderByIndices (targetList: TargetEntry list) orderByExpressions 
   let selectedExpressions = targetList |> List.map (fun targetEntry -> targetEntry.Expression)
   orderByExpressions |> List.map (mapOrderByIndex selectedExpressions)
 
+let private interpretDirection optDirection =
+  optDirection
+  |> Option.map (
+    function
+    | ParserTypes.Asc -> Ascending
+    | ParserTypes.Desc -> Descending
+    | _ -> failwith "Invalid `ORDER BY` clause"
+  )
+
+let private interpretNullsBehavior optNullsBehavior =
+  optNullsBehavior
+  |> Option.map (
+    function
+    | ParserTypes.NullsFirst -> NullsFirst
+    | ParserTypes.NullsLast -> NullsLast
+    | _ -> failwith "Invalid `ORDER BY` clause"
+  )
+
+let private interpretOrderByExpression rangeColumns expression =
+  match expression with
+  | ParserTypes.OrderSpec (expression, optDirection, optNullsBehavior) ->
+    let (direction, nullsBehavior) =
+      // we want the default nulls behavior to be "NULL values are largest", `ORDER BY x DESC` is a special case
+      match (interpretDirection optDirection, interpretNullsBehavior optNullsBehavior) with
+      | None, None -> Ascending, NullsLast
+      | Some Ascending, None -> Ascending, NullsLast
+      | Some Descending, None -> Descending, NullsFirst
+      | None, Some nullsBehavior -> Ascending, nullsBehavior
+      | Some direction, Some nullsBehavior -> direction, nullsBehavior
+
+    (mapExpression rangeColumns expression), direction, nullsBehavior
+  | _ -> failwith "Invalid `ORDER BY` clause"
+
 // ----------------------------------------------------------------
 // Query range
 // ----------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


cristianberneanu:
> I'm still scratching my head, if there are any anonymity-related concerns with ORDER BY, or things will "just work".

The anonymization mechanism is consistent with regards to data ordering, so `ORDER BY` clauses before the anonymization step should have no effect (but a test to make sure would be nice). And after anonymization the data is no longer sensitive, so it can be processed freely.

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -125,6 +125,56 @@ let private mapGroupByIndices (targetList: TargetEntry list) groupByExpressions
...
+  match expression with
+  | ParserTypes.OrderSpec (expression, optDirection, optNullsBehavior) ->
+    let (direction, nullsBehavior) =
+      // we want the default nulls behavior to be "NULL values are largest", `ORDER BY x DESC` is a special case
+      match (interpretDirection optDirection, interpretNullsBehavior optNullsBehavior) with
+      | None, None -> Ascending, NullsLast
+      | Some Ascending, None -> Ascending, NullsLast
+      | Some Descending, None -> Descending, NullsFirst
+      | None, Some nullsBehavior -> Ascending, nullsBehavior
+      | Some direction, Some nullsBehavior -> direction, nullsBehavior
```
cristianberneanu, line 173: I think this should be best handled during parsing.

src/OpenDiffix.Core/Parser.fs
```diff
@@ -91,6 +95,12 @@ module QueryParser =
 
   let limitClause = word "LIMIT" >>. puint32
 
+  let orderSpec =
+    expr .>>. opt (asc <|> desc) .>>. opt (nullsFirst <|> nullsLast) .>> spaces
+    |>> (fun ((expr, direction), nullsBehavior) -> Expression.OrderSpec(expr, direction, nullsBehavior))
```
cristianberneanu, line 100: This would be a good place to handle default behavior.

src/OpenDiffix.Core/ParserTypes.fs
```diff
@@ -39,6 +44,7 @@ and Expression =
   | Join of joinType: JoinType * left: Expression * right: Expression * on: Expression
   | SubQuery of subQuery: SelectQuery * alias: string
   | Function of functionName: string * Expression list
+  | OrderSpec of expr: Expression * direction: Expression option * nullsBehavior: Expression option
```
cristianberneanu, line 47: If default behavior is handled during parsing, there is no more need for the `option` attribute here.

pdobacz:
> > I'm still scratching my head, if there are any anonymity-related concerns with ORDER BY, or things will "just work".
> 
> The anonymization mechanism is consistent with regards to data ordering, so `ORDER BY` clauses before the anonymization step should have no effect (but a test to make sure would be nice). And after anonymization the data is no longer sensitive, so it can be processed freely.

So my naive thinking would be to attack like this:


```
dotnet run --project src/OpenDiffix.CLI/ -- -f ../data/taxi.sqlite --aid-columns taxi.hack -q "SELECT vendor_id, rate_code, sf_flag, COUNT(DISTINCT hack), pickup_datetime from taxi GROUP BY vendor_Id, rate_code, sf_flag, pickup_datetime ORDER BY 4;"
```

and hope for the ordering to be done not via the noisy `COUNT`s but via the raw ones. This doesn't seem to be the case - the `COUNT` column is ordered in the output so I guess we're using the anonymized counts :+1: . I'm not entirely sure **yet** which part of the code ensures this, but it seems all is in order.

cristianberneanu:
The anonymization step occurs during aggregation, while ordering is done afterwards.

cristianberneanu (approved):
👍

PR #281, opened 2021-11-04
Make bucket seed independent of grouping order.

cristianberneanu:


src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
```diff
@@ -6,26 +6,30 @@ open FsUnit.Xunit
...
     }
 
+  let noisyAnonParams = { AnonymizationParams.Default with TableSettings = tableSettings }
+
   let runQueryWithCustomAnonParams anonymizationParams query =
     let queryContext = QueryContext.make anonymizationParams db.DataProvider
     run queryContext query
 
-  let runQuery = runQueryWithCustomAnonParams anonParams
+  let runQuery = runQueryWithCustomAnonParams noislessAnonParams
```
pdobacz, line 32: `noiseless`

pdobacz (approved):
just a tiny typo and LGTM

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -403,11 +403,12 @@ let private computeNoiseLayers anonParams query =
...
-    |> Seq.collect (collectSeedMaterials rangeColumns)
-    |> String.join ","
-    |> System.Text.Encoding.UTF8.GetBytes
-    |> Hash.bytes
+    |> Seq.map (
+      normalizeNoiseLayerExpression
+      >> collectSeedMaterials rangeColumns
+      >> String.join ","
+    )
+    |> Hash.strings 0UL
```
edongashi, line 411: Could we end up in a situation where bucket labels cancel each other out because of the XOR?

cristianberneanu: I don't think this is a problem if it happens by chance for different expressions, since 0 is a seed value just as valid as any other. But it does happen when the same expression is used multiple times for grouping, cancelling its effect on the bucket seed. I will normalize this case in a future PR. Good catch!

pdobacz (approved):

PR #282, opened 2021-11-10
More grouping expressions normalizations.

### LINKED ISSUE #279: Normalize identity ranges for integers.
Expressions like `cast(column as real)` and `round/floor/ceil(column)` should produce the same SQL seed as a simple column selection for `integer` columns.

cristianberneanu:
Closes #279.

### DIFF
--- a/src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
@@ -226,4 +226,10 @@ type Tests(db: DBFixture) =
       "SELECT count(*) FROM customers_small GROUP BY round_by(age, 10), city ORDER BY 1"
       "SELECT count(*) FROM customers_small GROUP BY city, round_by(age, 10) ORDER BY 1"
 
+  [<Fact>]
+  let ``Duplicated grouping doesn't change results`` () =
+    equivalentQueries
+      "SELECT count(*) FROM customers_small GROUP BY city, city ORDER BY 1"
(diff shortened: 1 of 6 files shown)


src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
```diff
@@ -226,4 +226,10 @@ type Tests(db: DBFixture) =
       "SELECT count(*) FROM customers_small GROUP BY round_by(age, 10), city ORDER BY 1"
       "SELECT count(*) FROM customers_small GROUP BY city, round_by(age, 10) ORDER BY 1"
 
+  [<Fact>]
+  let ``Duplicated grouping doesn't change results`` () =
+    equivalentQueries
+      "SELECT count(*) FROM customers_small GROUP BY city, city ORDER BY 1"
+      "SELECT count(*) FROM customers_small GROUP BY city ORDER BY 1"
```
edongashi, line 233: Will the result also be identical? In postgres I noticed that duplicate labels are dropped.

edongashi (approved):

src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
cristianberneanu: I don't understand the question, can you please rephrase it?
This test asserts that the results are identical.

edongashi: Ah, I see. I thought it just tests for noise layers.

PR #287, opened 2021-11-23
Support aggregator merging

edongashi:


src/OpenDiffix.Core.Tests/Aggregator.Tests.fs
```diff
@@ -0,0 +1,90 @@
...
+
+open CommonTypes
+
+let diffixCount = AggregateFunction(DiffixCount, { AggregateOptions.Default with Distinct = false })
+let distinctDiffixCount = AggregateFunction(DiffixCount, { AggregateOptions.Default with Distinct = true })
+let diffixLowCount = AggregateFunction(DiffixLowCount, AggregateOptions.Default)
+
+let randomNullableInteger (random: System.Random) =
+  let x = random.Next(100) - 10 // ~10% chance to get a NULL
+  if x < 1 then Null else Integer(int64 x)
```
cristianberneanu, line 14: This seems a bit convoluted. Alternatively:
```FS
  if random.Next(10) = 0 then Null else Integer(int64 random.Next(100))
```

cristianberneanu (dismissed):
👍 
The `DiffixLowCount` aggregator has special importance and it is needed in many different cases. We should consider making it always present implicitly, either by creating a bucket structure with it as a mandatory field, or by forcing `aggregators[0]` to always be `DiffixLowCount`.

src/OpenDiffix.Core.Tests/Aggregator.Tests.fs
edongashi: Yeah, looks better!

edongashi:
> The DiffixLowCount aggregator has special importance and it is needed in many different cases. We should consider making it always present implicitly

I was thinking similarly of some canonical `Bucket` state during aggregation. Let's see what it needs in practice when implementing the simplified hook.

cristianberneanu (approved):

PR #288, opened 2021-11-23
Fix thread unsafe hash calculation

### LINKED ISSUE #286: Nondeterministic noise in tests
Any idea why aggregation produces different noise samples across test runs? This causes flaky tests unless you configure all noise to be zero. 

Also `OpenDiffix.Core.QueryEngineTests+Tests.Duplicated grouping doesn't change results` fails sometimes...
```
Expected: Equals { Columns = [{ Name = "count"
               Type = IntegerType }]
  Rows = [[|Integer 7L|]; [|Integer 10L|]] }
Actual:   { Columns = [{ Name = "count"
               Type = IntegerType }]
  Rows = [[|Integer 10L|]; [|Integer 10L|]] }
```

edongashi:
Closes #286. This caused different hashes because the hash algorithm is apparently stateful.

### DIFF
--- a/src/OpenDiffix.Core/Anonymizer.fs
+++ b/src/OpenDiffix.Core/Anonymizer.fs
@@ -29,9 +29,8 @@ let private randomNormal stdDev (seed: Hash) =
   let randomNormal = Math.Sqrt(-2.0 * log u1) * Math.Sin(2.0 * Math.PI * u2)
   stdDev * randomNormal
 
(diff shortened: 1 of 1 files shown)


cristianberneanu (approved):
Nice catch! Sadly, I always run the tests single-threaded, as a workaround to some unrelated .NET issue under WSL, so I never encountered this...

PR #290, opened 2021-11-25
Add aggregation hook

### LINKED ISSUE #289: Standardize a model for buckets and post aggregation logic

edongashi:
Closes #289.

Executor hook will be removed once we migrate LED.

### DIFF
--- a/src/OpenDiffix.Core/CommonTypes.fs
+++ b/src/OpenDiffix.Core/CommonTypes.fs
@@ -166,10 +166,13 @@ type AnonymizationParams =
 
 type ExecutorHook = ExecutionContext -> Plan -> seq<Row>
 
+type PostAggregationHook = AggregationContext -> seq<AggregationBucket> -> seq<AggregationBucket>
+
 type QueryContext =
   {
     AnonymizationParams: AnonymizationParams
     DataProvider: IDataProvider
+    PostAggregationHook: PostAggregationHook
     ExecutorHook: ExecutorHook option
   }
 
@@ -195,6 +198,26 @@ type Plan =
 // ----------------------------------------------------------------
 // Executor
 // ----------------------------------------------------------------
+
+type IAggregator =
+  abstract Transition : Value list -> unit
+  abstract Merge : IAggregator -> unit
+  abstract Final : ExecutionContext -> Value
+
+type AggregationContext =
+  {
+    BaseExecutionContext: ExecutionContext
+    GroupingLabels: Expression array
+    AggregatorFunctions: Function array
+    AggregatorArgs: array<Expression list>
+  }
+
+type AggregationBucket =
+  {
+    Group: Row
+    Aggregators: IAggregator array
+    ExecutionContext: ExecutionContext
+  }
 
 type NoiseLayers =
   {
@@ -312,6 +335,7 @@ module QueryContext =
     {
       AnonymizationParams = anonParams
       DataProvider = dataProvider
+      PostAggregationHook = fun _aggregationContext -> id
       ExecutorHook = None
     }
 
--- a/src/OpenDiffix.Core/CommonTypes.fs
+++ b/src/OpenDiffix.Core/CommonTypes.fs
@@ -209,7 +209,6 @@ type AggregationContext =
     ExecutionContext: ExecutionContext
     GroupingLabels: Expression array
     Aggregators: (Function * Expression list) array
-    LowCountIndex: int option
   }
 
 type AggregationBucket =
--- a/src/OpenDiffix.Core/CommonTypes.fs
+++ b/src/OpenDiffix.Core/CommonTypes.fs
@@ -217,6 +217,7 @@ type AggregationBucket =
     Group: Row
     Aggregators: IAggregator array
     ExecutionContext: ExecutionContext
+    LowCount: bool
   }
 
 type NoiseLayers =
--- a/src/OpenDiffix.Core/CommonTypes.fs
+++ b/src/OpenDiffix.Core/CommonTypes.fs
@@ -206,10 +206,10 @@ type IAggregator =
 
 type AggregationContext =
   {
-    BaseExecutionContext: ExecutionContext
+    ExecutionContext: ExecutionContext
     GroupingLabels: Expression array
-    AggregatorFunctions: Function array
-    AggregatorArgs: array<Expression list>
+    Aggregators: (Function * Expression list) array
+    LowCountIndex: int option
   }
 
 type AggregationBucket =
@@ -217,7 +217,6 @@ type AggregationBucket =
     Group: Row
     Aggregators: IAggregator array
     ExecutionContext: ExecutionContext
-    LowCount: bool
   }
 
 type NoiseLayers =
--- a/src/OpenDiffix.Core/Executor.fs
+++ b/src/OpenDiffix.Core/Executor.fs
@@ -59,22 +59,11 @@ let private executeAggregate executionContext (childPlan, groupingLabels, aggreg
   let isGlobal = Array.isEmpty groupingLabels
   let aggFns, aggArgs = aggregators |> Array.ofList |> Utils.unpackAggregators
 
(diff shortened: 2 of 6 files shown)


src/OpenDiffix.Core/CommonTypes.fs
```diff
@@ -196,6 +199,26 @@ type Plan =
...
+  abstract Transition : Value list -> unit
+  abstract Merge : IAggregator -> unit
+  abstract Final : ExecutionContext -> Value
+
+type AggregationContext =
+  {
+    BaseExecutionContext: ExecutionContext
+    GroupingLabels: Expression array
+    AggregatorFunctions: Function array
+    AggregatorArgs: array<Expression list>
```
cristianberneanu, line 212: These should probably be grouped together.
```fs
 Aggregators: (Function, Expression list) array
```

src/OpenDiffix.Core/CommonTypes.fs
```diff
@@ -196,6 +199,26 @@ type Plan =
...
 // ----------------------------------------------------------------
 
+type IAggregator =
+  abstract Transition : Value list -> unit
+  abstract Merge : IAggregator -> unit
+  abstract Final : ExecutionContext -> Value
+
+type AggregationContext =
+  {
+    BaseExecutionContext: ExecutionContext
```
cristianberneanu, line 209: The `Base` tag doesn't seem that useful.

src/OpenDiffix.Core/CommonTypes.fs
```diff
@@ -166,10 +166,13 @@ type AnonymizationParams =
 
 type ExecutorHook = ExecutionContext -> Plan -> seq<Row>
 
+type PostAggregationHook = AggregationContext -> seq<AggregationBucket> -> seq<AggregationBucket>
+
 type QueryContext =
   {
     AnonymizationParams: AnonymizationParams
     DataProvider: IDataProvider
+    PostAggregationHook: PostAggregationHook
```
cristianberneanu, line 175: What do you think about making this a list of hooks instead?
It would then be easier to keep the code for LED and *-bucket separate.
Or maybe some experimentation is needed first...

src/OpenDiffix.Core/CommonTypes.fs
```diff
@@ -217,6 +217,7 @@ type AggregationBucket =
     Group: Row
     Aggregators: IAggregator array
     ExecutionContext: ExecutionContext
+    LowCount: bool
```
cristianberneanu, line 220: I think this field should be an `IAggregator` object. That way, the low-count property can be both evaluated and merged with another bucket's property.

src/OpenDiffix.Core/Executor.fs
```diff
@@ -96,7 +107,19 @@ let private executeAggregate executionContext (childPlan, groupingLabels, aggreg
...
+    { bucket with
+        // A bucket is low count if any low count aggregator in scope returns true.
+        LowCount =
+          lowCountAggs
+          |> Array.exists (fun aggIndex ->
+            bucket.Aggregators.[aggIndex].Final bucket.ExecutionContext
+            |> Value.unwrapBoolean
+          )
+    }
+  )
```
cristianberneanu, line 122: I don't think this approach is optimal.
I would prefer if we uniquely set the low-count aggregator index during query analysis. Or if we implicitly make it to always be the first in the list. Then we can drop it from our custom queries in the desktop client.

cristianberneanu (approved):

src/OpenDiffix.Core/CommonTypes.fs
edongashi: I think that can be done with a simple composition in client code. If not, we change it later.

src/OpenDiffix.Core/Executor.fs
edongashi: > Or if we implicitly make it to always be the first in the list.

How do we handle standard queries if we do this?

> Then we can drop it from our custom queries in the desktop client.

How do we pass AID information? If through table config, then queries will be rewritten which we don't want.

src/OpenDiffix.Core/CommonTypes.fs
edongashi: The `IAggregator` can be found in `Aggregators`. We may pass an index or make it first like you suggested.
The reason for the constant boolean is that for LED we don't want a change in `LowCount` mid-way because it will make the algorithm order dependent, although extremely unlikely to be reproducible for an attack.

cristianberneanu: This is a concern specific to the LED hook, which shouldn't spill into the overall design.
Will it be too complex to remember the initial atribute value inside the hook? If yes, then keep this as is.

edongashi: Ok, it can be memo'd inside the LED hook, not a problem. I'm still thinking on how do we treat multiple LCF's in scope.

Maybe we raise during compilation?

src/OpenDiffix.Core/Executor.fs
cristianberneanu: > How do we handle standard queries if we do this?

Either we always compute it and then ignore it, or this only applies to anonymized queries. Whatever generates less special cases.

> How do we pass AID information?

You're right, if we mark columns as AIDs then we need to add a toggle for query rewriting. Not ideal.
Alternatively, we can run the query twice, once as standard and once as anonymized, then merge the results. Also not ideal.
This requires more thought... For now, I think it is still worth it to put the index in the execution context and set it during the analysis stage.

src/OpenDiffix.Core/CommonTypes.fs
cristianberneanu: I don't currently see the utility of a scenario in which we can have more than one, so raising would be fine for now.

edongashi: No utility whatsoever, but we limit the freehand queries that can be called via reference CLI.

src/OpenDiffix.Core/Executor.fs
edongashi: Passing an `int option` for now. We can skip the hooks if there is no LCF - meaning the query is likely standard, or this in-between query from desktop which is technically anonymizing but written manually...

src/OpenDiffix.Core/CommonTypes.fs
```diff
@@ -206,18 +206,17 @@ type IAggregator =
 
 type AggregationContext =
   {
-    BaseExecutionContext: ExecutionContext
+    ExecutionContext: ExecutionContext
     GroupingLabels: Expression array
-    AggregatorFunctions: Function array
-    AggregatorArgs: array<Expression list>
+    Aggregators: (Function * Expression list) array
+    LowCountIndex: int option
```
cristianberneanu, line 212: I meant that this should be put in the execution context so it doesn't have to be computed during aggregation.
Or is that more convoluted to do?

cristianberneanu (dismissed):

src/OpenDiffix.Core/CommonTypes.fs
edongashi: The execution context is query (request) wide, whereas the aggregation context is node-specific. I think it's cleaner this way as it holds state more locally.

edongashi: Alternatively we can store it in the plan node. But this is a computed value from aggFns so it's redundant to carry around.

cristianberneanu: > The execution context is query (request) wide

Good point!

edongashi: Or we don't pass it at all! The hooks can find it easily...

cristianberneanu: Good idea! Even better to provide a utility function for that.

cristianberneanu (approved):

PR #293, opened 2021-12-01
Document IAggregator and QueryEngine.run a bit

pdobacz:
I was going through `reference` and added some auxiliary comments, which would have made sense for me. Let me know if they're accurate and make sense to have.

### DIFF
--- a/src/OpenDiffix.Core/QueryEngine.fs
+++ b/src/OpenDiffix.Core/QueryEngine.fs
@@ -15,6 +15,33 @@ let rec private extractColumns query =
 
 type QueryResult = { Columns: Column list; Rows: Row list }
 
+// The main entrypoint to the `OpenDiffix.Core` library.
+//
+// Takes in a SQL `statement` as main parameter along with a set of parameters in `queryContext`. Executes the query
+// against the `IDataProvider` provided in the `queryContext` and returns a structure representing the result table. The
+// data in the result table will be anonymized (or will contain facilities necessary to anonymize), according to the
+// modes of operation listed below.
+//
+// Supports two main modes of operation:
+// 1. (anonymizing query) Standard SQL statement, with `aidColumns` specified within the `queryContext`
+//
+// Rewrites the statement, so that the query result is anonymized according to the `aidColumns` provided. For example,
+// it appends `AND NOT diffix_low_count(aidColumns)` to the `HAVING` SQL clause and rewrites regular SQL aggregators to
+// Diffix aggregators (e.g. `count` -> `diffix_count`, see below).
+//
+// 2. (standard query) SQL statement with Diffix aggregators, no `aidColumns` specified within the `queryContext`
+//
+// (**NOTE**) some AID columns may (and should) be used within the Diffix aggregator functions.
+//
+// Does not rewrite the statement. The caller is expected to use Diffix aggregator functions manually, in order to
+// anonymize the query result. **NOTE** that the Diffix aggregators are optional, if they're not used, the result is the
+// same as for a regular SQL query.
+//
+// Diffix aggregator functions available:
+//   - `diffix_low_count(aid_columns_expression)` - `true` if the bucket has low count and should be suppressed
+//   - `diffix_count(aid_columns_expression)` - anonymized count of the bucket
+//
+// `aid_columns_expression` should be a list of `aid_column` or `DISTINCT aid_column`.
 let run queryContext statement : QueryResult =
   let query, noiseLayers =
     statement
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ around the complex and large codebase of a real world database engine.
 The codebase is currently organized in a number of projects:
 
 - `OpenDiffix.Core`: Contains the meat of this project. It is the query and anonymization engine.
+  The main public interface to it is the [`QueryEngine.run` function here](src/OpenDiffix.Core/QueryEngine.fs)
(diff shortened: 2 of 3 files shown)


src/OpenDiffix.Core/QueryEngine.fs
```diff
@@ -15,6 +15,32 @@ let rec private extractColumns query =
 
 type QueryResult = { Columns: Column list; Rows: Row list }
 
+// The main entrypoint to the `OpenDiffix.Core` library.
+// Takes in a SQL `statement` as main parameter along with a set of
+// parameters in `queryContext`.
+//
+// Returns the query result as a list of data rows, along with the
+// columns specifications.
```
edongashi, line 23: This is quite obvious a few lines above with `type QueryResult = { Columns: Column list; Rows: Row list }`

src/OpenDiffix.Core/QueryEngine.fs
```diff
@@ -15,6 +15,32 @@ let rec private extractColumns query =
...
+//
+// Supports two main modes of operation:
+// I. Standard SQL statement, with `aidColumns` specified within the `queryContext`
+//
+// Will rewrite the statement so that the query result is anonymized according to the
+// `aidColumns` provided. For example, it will append `AND NOT diffix_low_count(aidColumns)`
+// to the `HAVING` SQL clause and rewrite regural SQL aggregators to Diffix aggregators
```
edongashi, line 30: regural -> regular

src/OpenDiffix.Core/QueryEngine.fs
```diff
@@ -15,6 +15,32 @@ let rec private extractColumns query =
...
+// (e.g. `count` -> `diffix_low_count`, see below)
+//
+// II. SQL statement with Diffix aggregators, no `aidColumns` specified within the `queryContext`
```
edongashi, line 33: Can also be without any custom aggregators i.e. regular query.

src/OpenDiffix.Core/QueryEngine.fs
```diff
@@ -15,6 +15,32 @@ let rec private extractColumns query =
...
+//
+// (**NOTE**) some AID columns may (and should) be used within the Diffix aggregator functions
+//
+// Will not rewrite the statement. The caller is expected to use Diffix aggregator functions
+// manually, in order to anonymize the query result manually.
```
edongashi, line 38: Doesn't add too much information

edongashi:
The comments in the runner may as well be added to root `README`, which is outdated anyway and needs some cleaning up. What do you think?

src/OpenDiffix.Core/QueryEngine.fs
pdobacz: hm, this is in the vein that "it does that, and doesn't do anything else or return any intermediate result, which just happens to be typed `Row list`, plus `Row` is plain-old-data", which isn't that obvious to anyone doing a first read. I can try to rephrase to emphasis that.

src/OpenDiffix.Core/QueryEngine.fs
pdobacz: hm, just noticed vscode spellchecker extension doesn't work for `.fs` files, sorry about that

src/OpenDiffix.Core/QueryEngine.fs
pdobacz: which part exactly? The two operation modes where very confusing to me, when I was first confronted with this. This distinction is important to make in my opinion, to capture the somewhat "leaky" abstraction, which is OK to have, but confusing, if not made explicit.

My aim is to make the answer to the question "which of the components is responsible for the anonymization?" hard to miss

pdobacz:
> The comments in the runner may as well be added to root `README`, which is outdated anyway and needs some cleaning up. What do you think?

Good idea, I'll try to draw a line somewhere and coordinate these docs

src/OpenDiffix.Core/QueryEngine.fs
edongashi: I see what you mean. We used to call these standard queries and anonymizing queries. OK with you if I tweak the comments a bit to match with our terminology?

src/OpenDiffix.Core/QueryEngine.fs
pdobacz: technically true, but then it doesn't make sense to use it, hence not a "main mode of operation". I'll make it clearer that here a "mode of operation that brings the user closer to anonymous results this way or another" is meant.

src/OpenDiffix.Core/QueryEngine.fs
pdobacz: hm, calling these `standard` vs `anonymizing` queries doesn't fully make it clear. I want to document the intended use of this function (or the entire lib, since this is its entrypoint) to anyone without relying to terminology coined elsewhere, ELI5 style. But I'm working on a significant revamp, I hope it will read better then

pdobacz:
> > The comments in the runner may as well be added to root `README`, which is outdated anyway and needs some cleaning up. What do you think?
> 
> Good idea, I'll try to draw a line somewhere and coordinate these docs

Actually, I looked at README.md and there isn't any overlapping docs there, so I just pointed the reader to the docs about the entrypoint. I think it's good to keep the library docs at the code level, so that they're easily accessible to anyone who ends up arriving from the `CLI`, `Service` or tests.

I couldn't otherwise find anything off or outdated in the README.md, I think it does what it should do.

edongashi (approved):

PR #298, opened 2021-12-09
Keep track of row count in buckets

edongashi:
This will also be useful to get a summary without the result iteration that desktop does.

### DIFF
--- a/src/OpenDiffix.Core/CommonTypes.fs
+++ b/src/OpenDiffix.Core/CommonTypes.fs
@@ -232,6 +232,7 @@ type AggregationContext =
 type Bucket =
   {
     Group: Row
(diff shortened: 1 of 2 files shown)


pdobacz (approved):
would `RowCount` fit under the `Attributes`, so that we keep all these "useful data" under one roof?

I'm not sure myself - what becomes a "fixed" attribute like `RowCount` and what goes under `Attributes`, do we have a criteria for this?

edongashi:
I think `RowCount` is a pretty natural property of `Bucket`. It indicates how many times Aggregators have been transitioned and is also useful for debugging.

As for Attributes, those are more for plugins where we don't know the keys beforehand.

PR #305, opened 2022-01-10
Update ORDER BY to match that of PostgreSQL

pdobacz:
Split out of the property based testing spike. This is the only change to the `reference` implementation required to make these tests pass.

### DIFF
--- a/src/OpenDiffix.Core.Tests/Value.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/Value.Tests.fs
@@ -44,3 +44,21 @@ module ComparerTests =
     [ Integer 10L; Null; Integer 1L; Null; Integer 5L ]
     |> sort Descending NullsLast
     |> should equal [ Integer 10L; Integer 5L; Integer 1L; Null; Null ]
+
+  [<Fact>]
+  let ``String sort groups by letter case`` () =
+    [ String "a"; String "C"; String "A"; String "c" ]
+    |> sort Ascending NullsLast
+    |> should equal [ String "a"; String "A"; String "c"; String "C" ]
+
+  [<Fact>]
+  let ``String sort ignores whitespace and symbols`` () =
+    [ String " a"; String "  C"; String "+  A"; String "   c" ]
+    |> sort Ascending NullsLast
+    |> should equal [ String " a"; String "+  A"; String "   c"; String "  C" ]
+
+  [<Fact>]
+  let ``String sort uses whitespace and symbols comparison on conflict`` () =
+    [ String "+a"; String "a"; String " a"; String "b" ]
+    |> sort Ascending NullsLast
+    |> should equal [ String " a"; String "+a"; String "a"; String "b" ]
--- a/src/OpenDiffix.Core/Expression.fs
+++ b/src/OpenDiffix.Core/Expression.fs
@@ -220,18 +220,22 @@ let rec evaluate (row: Row) (expr: Expression) =
 
 /// Sorts a row sequence based on the given orderings.
 let sortRows orderings (rows: Row seq) =
-  let rec compare orderings rowA rowB =
-    match orderings with
+  let exprsComparers =
+    orderings
+    |> List.map (fun (OrderBy (expr, direction, nulls)) -> expr, Value.comparer direction nulls)
+
+  let rec compare exprsComparers rowA rowB =
+    match exprsComparers with
     | [] -> 0
-    | OrderBy (expr, direction, nulls) :: tail ->
+    | (expr, comparer) :: tail ->
       let valueA = evaluate rowA expr
       let valueB = evaluate rowB expr
 
-      match Value.comparer direction nulls valueA valueB with
+      match comparer valueA valueB with
       | 0 -> compare tail rowA rowB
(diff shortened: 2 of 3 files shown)


src/OpenDiffix.Core/Value.fs
```diff
@@ -46,6 +47,20 @@ let comparer direction nulls =
...
+      // so we ignore first.
+      let comparisonIgnoreSymbols =
+        directionValue
+        * compareInfo.Compare(x, y, CompareOptions.IgnoreSymbols + CompareOptions.StringSort)
+      // if the former gives a tie, we include symbols.  `StringSort` means they come last and we group letter cases together
+      // reference: https://wiki.postgresql.org/wiki/FAQ#Why_do_my_strings_sort_incorrectly.3F
+      if (comparisonIgnoreSymbols <> 0) then
+        comparisonIgnoreSymbols
+      else
+        directionValue * compareInfo.Compare(x, y, CompareOptions.StringSort)
```
edongashi, line 63: Can we memo this so that it's outside the lambda? In the callsite I'd then memo the func instead of creating a new one for each sorting pair (may be more expensive).

src/OpenDiffix.Core.Tests/Value.Tests.fs
```diff
@@ -44,3 +44,21 @@ module ComparerTests =
     [ Integer 10L; Null; Integer 1L; Null; Integer 5L ]
     |> sort Descending NullsLast
     |> should equal [ Integer 10L; Integer 5L; Integer 1L; Null; Null ]
+
+  [<Fact>]
+  let ``String sort groups by letter case`` () =
+    [ String "a"; String "C"; String "A"; String "c" ]
+    |> sort Ascending NullsLast
+    |> should equal [ String "a"; String "A"; String "c"; String "C" ]
```
cristianberneanu, line 52: I am confused by this test (which I guess should belong in the previous commit) and the previous commit title.
The sort should be case insensitive, so the expected order should be `aACc`, but it is not.

src/OpenDiffix.Core/Expression.fs
```diff
@@ -220,18 +220,22 @@ let rec evaluate (row: Row) (expr: Expression) =
...
+  let rec compare exprsComparers rowA rowB =
+    match exprsComparers with
     | [] -> 0
-    | OrderBy (expr, direction, nulls) :: tail ->
+    | (expr, comparer) :: tail ->
       let valueA = evaluate rowA expr
       let valueB = evaluate rowB expr
 
-      match Value.comparer direction nulls valueA valueB with
+      match comparer valueA valueB with
```
cristianberneanu, line 234: Hm, this seems to be just a refactoring, not an optimization: the closure is still instantiated each time (and has to be), isn't it?

cristianberneanu (approved):

cristianberneanu:
> the property based testing spike

What do you mean by this? I never heard this term before, but, from what I understand, spike testing is related to performance testing, not property testing.

src/OpenDiffix.Core/Expression.fs
edongashi: `Value.comparer direction nulls` computes some constants before returning a comparer function.

`(Value.comparer direction nulls) -> returns func -> called with valueA valueB`

pdobacz:
> > the property based testing spike
> 
> What do you mean by this? I never heard this term before, but, from what I understand, spike testing is related to performance testing, not property testing.

I meant the property based testing [**spike**](https://en.wikipedia.org/wiki/Spike_(software_development)) in this meaning. I never heard the term "spike testing" before, so the entire thing turned out quite confusing indeed, apologies for this.

More concretely, I mean the PR is split out of #302, where I originally included the commits

src/OpenDiffix.Core/Value.fs
```diff
@@ -46,6 +52,19 @@ let comparer direction nulls =
...
     | _, Null -> -nullsValue
+    | String x, String y ->
+      // Using PostgreSQL string comparison as a template.
+      // https://wiki.postgresql.org/wiki/FAQ#Why_do_my_strings_sort_incorrectly.3F
+
+      // We want whitespace & punctuation comparison ("symbols" in .NET) to have smaller priority,
+      // so we ignore them first.
+      let comparisonIgnoreSymbols = directionValue * stringCompareInfo.Compare(x, y, ignoreSymbolsFlag)
+      // If the former gives a tie, we include symbols.
+      if (comparisonIgnoreSymbols <> 0) then
```
edongashi, line 63: Redundant parens?

src/OpenDiffix.Core.Tests/Value.Tests.fs
pdobacz: My bad, the tests ended up committed once, for the bulk change, I was doing a little bit of back and forth here.

On the expected order, I think we need the sort to be in the order, where the case is a secondary sort criteria. So not precisely "case insensitive" as I described the commit. Case insensitive would mean `aACc` indeed in this case, I made a mental shortcut, not thinking about it's not the same thing.

This way or another, the test is in line with what PostgreSQL does and what we should be doing.

PR #306, opened 2022-01-12
Ensure that generate.fsx generates same CSV as SQLite

pdobacz:
Split from #302 - the only commit from there which belongs in `reference`

### DIFF
--- a/data/generate.fsx
+++ b/data/generate.fsx
@@ -43,7 +43,11 @@ type Table =
     StaticRows: Field list list
   }
 
-let gRNG = System.Random(123) // Fixed seed because we want constant values
+// Fixed seed because we want constant values.
+// Mutable because we want to easily reset and generate identical tables in CSV.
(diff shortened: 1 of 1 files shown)


edongashi (approved):

data/generate.fsx
```diff
@@ -344,6 +348,10 @@ File.Delete(csvFilePath)
 
 printfn "Creating table %A as CSV" customers.Name
 
+// reset the RNG, and be careful about the same order of generated tables,
```
edongashi, line 351: I find these comments a little overly verbose.

pdobacz: Indeed, they are verbose :(. But this still beats missing this and testing on different data, which I did for a while :) . I think the TODO above, to make the generation of SQLite/CSV consistently seeded, is the proper way to resolve this. Or, ideally, one should be able to export CSV from SQLite, but I couldn't find a way which would play nicely with this generator. Maybe I could re-approach this later, with a fresh mind.

PR #308, opened 2022-01-18
Fix LCF threshold mean to be LowMeanGap*SD away from LowThreshold

pdobacz:


src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -63,7 +63,11 @@ let isLowCount (executionContext: ExecutionContext) (aidSets: HashSet<AidHash> s
         [ executionContext.NoiseLayers.BucketSeed; seedFromAidSet aidSet ]
         |> generateNoise anonParams.Salt "suppress" anonParams.Suppression.LayerSD
 
-      let thresholdMean = anonParams.Suppression.LowMeanGap + float anonParams.Suppression.LowThreshold
+      // `LowMeanGap` is the number of standard deviations between `LowThreshold` and desired mean
+      let thresholdMean =
+        anonParams.Suppression.LowMeanGap * anonParams.Suppression.LayerSD
```
cristianberneanu, line 68: We have two noise layers, so `TotalSD = LayerSD * sqrt(2)`.

cristianberneanu (changes requested):

src/OpenDiffix.Core/Anonymizer.fs
pdobacz: Right. I think in this case, I'll also make it a little more explicit in the parameter docs. It's easy to miss the `layer` and expect that `LayerSD` is the `standard deviation` meant here

pdobacz:
I took the liberty to force push the fixup, since it was a one-liner originally.

cristianberneanu (approved):

PR #309, opened 2022-01-20
Updates to rounding and count distinct to match PostgreSQL

pdobacz:
This is another PR facilitating the https://github.com/diffix/pg_diffix/issues/156 testing. We aim at making the `reference` implementation more in-line with what we can have on the `pg_diffix` end, in order to be able to compare the results.

This might or might not be the right path forward. These changes are induced by how PostgreSQL is implemented and by the need to have exact same results for same queries (to compare them using tools like property tests _easily_).

If we say this is not desired, and we won't be able to sustainably keep the implementations in-sync to this degree of accuracy, we have the following alternatives:
1. give up property testing (I think quite bad, it yields good results so far and makes navigating the discrepancies much easier)
2. make the property tests compare results with some leeway - unfortunately, there are cases where the system is inherently "unstable" and small difference (e.g. off-by-one difference in contribution counting) leads to big differences in the final query outcome.

EDIT: it turned out that one (most obtrusive) change that this PR originally included wasn't necessary, so the above disclaimers are not that relevant anymore.

Review in parallel with https://github.com/diffix/pg_diffix/pull/169


src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -199,14 +201,40 @@ let private distributeValues (valuesByAID: seq<AidHash * array<Value>>) : seq<Ai
...
+let private valueComparer = Value.comparer Ascending NullsFirst
+
+let private comparePerAidContributions
+  ((aidA, valuesA): AidHash * Value array)
+  ((aidB, valuesB): AidHash * Value array)
+  =
+  if valuesA.Length <> valuesB.Length then
+    valuesA.Length - valuesB.Length
+  else
+    let result =
```
pdobacz, line 213: the new sorter is different by that it sorts using 3 criteria:
1. array length
2. (new!) first differing array element
3. aid

This makes the sorting more cross-platform, as we don't get sorted differently by different hashing algorithms. If the hashing algorithms are made equal, we can get rid of this change.

src/OpenDiffix.Core/Utils.fs
```diff
@@ -69,3 +69,8 @@ module Hash =
 
   let strings start data =
     data |> Seq.distinct |> Seq.map string |> Seq.fold (^^^) start
+
+// introduced for compatibility of math-related functions with C (PostgreSQL)
+module CMath =
```
edongashi, line 74: I would rename this module to `Math` (F# allows module merging) and the function to `roundAwayFromZero` or  something similar.

pdobacz: Sounds good. This has a slight drawback of being longer and chopping up some lines in `Expressions.fs` (even if we `open` it) but :shrug:

src/OpenDiffix.Core/Anonymizer.fs
pdobacz: NOTE: consistent hashing allowed us to get rid of this change, hence comment is `Outdated`.

cristianberneanu (approved):

PR #311, opened 2022-01-26
Suggestions to refactor the Anonymizer

pdobacz:
A collection of 4 independent refactors, I've put comments into the commits where necessary.

### DIFF
--- a/src/OpenDiffix.Core/Aggregator.fs
+++ b/src/OpenDiffix.Core/Aggregator.fs
@@ -167,9 +167,8 @@ type private DiffixCount(minCount) =
         Integer minCount
       else
         match Anonymizer.count executionContext state with
-        | Null -> Integer minCount
-        | Integer value -> Integer(max value minCount)
-        | value -> value
+        | Anonymizer.Result.NotEnoughAIDVs -> Integer minCount
+        | Anonymizer.Result.Count value -> Integer(max value minCount)
 
 type private DiffixCountDistinct(minCount) =
   let mutable aidsCount = 0
@@ -218,9 +217,8 @@ type private DiffixCountDistinct(minCount) =
 
     member this.Final executionContext =
       match Anonymizer.countDistinct executionContext aidsCount aidsPerValue with
-      | Null -> Integer minCount
-      | Integer value -> Integer(max value minCount)
-      | value -> value
+      | Anonymizer.Result.NotEnoughAIDVs -> Integer minCount
+      | Anonymizer.Result.Count value -> Integer(max value minCount)
 
 type private DiffixLowCount() =
   let mutable state: HashSet<AidHash> [] = null
--- a/src/OpenDiffix.Core/Aggregator.fs
+++ b/src/OpenDiffix.Core/Aggregator.fs
@@ -167,8 +167,8 @@ type private DiffixCount(minCount) =
         Integer minCount
       else
         match Anonymizer.count executionContext state with
-        | Anonymizer.Result.NotEnoughAIDVs -> Integer minCount
-        | Anonymizer.Result.Count value -> Integer(max value minCount)
+        | Anonymizer.CountResult.NotEnoughAIDVs -> Integer minCount
+        | Anonymizer.CountResult.Ok value -> Integer(max value minCount)
 
 type private DiffixCountDistinct(minCount) =
   let mutable aidsCount = Option<int>.None
@@ -220,8 +220,8 @@ type private DiffixCountDistinct(minCount) =
         Integer minCount
       else
         match Anonymizer.countDistinct executionContext aidsCount.Value aidsPerValue with
-        | Anonymizer.Result.NotEnoughAIDVs -> Integer minCount
-        | Anonymizer.Result.Count value -> Integer(max value minCount)
+        | Anonymizer.CountResult.NotEnoughAIDVs -> Integer minCount
+        | Anonymizer.CountResult.Ok value -> Integer(max value minCount)
 
 type private DiffixLowCount() =
   let mutable state: HashSet<AidHash> [] = null
--- a/src/OpenDiffix.Core/Aggregator.fs
+++ b/src/OpenDiffix.Core/Aggregator.fs
@@ -171,7 +171,7 @@ type private DiffixCount(minCount) =
         | Anonymizer.Result.Count value -> Integer(max value minCount)
 
 type private DiffixCountDistinct(minCount) =
-  let mutable aidsCount = 0
+  let mutable aidsCount = Option<int>.None
   let aidsPerValue = Dictionary<Value, HashSet<AidHash> array>()
 
   member this.AidsCount = aidsCount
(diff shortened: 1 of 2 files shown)


src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -247,6 +247,10 @@ let isLowCount (executionContext: ExecutionContext) (aidSets: HashSet<AidHash> s
   )
   |> Seq.reduce (||)
 
+type Result =
```
pdobacz, line 249: I have a problem with naming here, suggestions welcome. 

In general my aim here is for the types to better represent the meaning. `Value` isn't good, because it can (completely nonsensically here) be a `String`.

A "last resort" alternative is `Option<int>`, but it is less meaningful, why the `None` in here is better described by the verbose `NotEnoughAIDVs`. I'd like this to be somehow captured, and set apart from the other "optionalities" involved around the code. `NotEnoughAIDVs` communicates to the `Aggregator` what needs to be done.

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -171,7 +171,7 @@ type private DiffixCount(minCount) =
         | Anonymizer.Result.Count value -> Integer(max value minCount)
 
 type private DiffixCountDistinct(minCount) =
-  let mutable aidsCount = 0
+  let mutable aidsCount = Option<int>.None
```
pdobacz, line 174: My problem with `0` is that it conflates:
1. an invalid situation of 0 AID instances
2. non-initialized instance of `DiffixCountDistinct`

this particular change clarifies this, and makes the check in `Final` cleaner.

src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -283,8 +282,11 @@ let countDistinct
     byAid
     |> List.choose id
     |> anonymizedSum
-    |> Option.defaultValue 0.
-    |> (Math.roundAwayFromZero >> int64 >> (+) safeCount >> max 0L >> Result.Count)
+    |> Math.roundAwayFromZero
+    |> int64
```
pdobacz, line 286: I was wondering if I could move these 2 lines to `anonymizedSum`, but doesn't the latter also cater for float results, like, coming from a `SUM` aggregator?

pdobacz: also, I now realize that the combination of `()`s and `>>`s was here to force it to a single line, pls confirm I'll undo this change

src/OpenDiffix.Core/Anonymizer.fs
edongashi: How about `CountResult`? `Result` is a default type and I wouldn't prefer creating ambiguity.

src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -274,13 +278,13 @@ let countDistinct
   // we can only report the count of values we already know to be safe as they
   // individually passed low count filtering.
   if byAid |> List.exists ((=) None) then
-    if safeCount > 0L then Integer safeCount else Null
+    if safeCount > 0L then safeCount |> Result.Count else NotEnoughAIDVs
```
edongashi, line 280: Pipe can be removed so that it reads more straightforward: `Result.Count safeCount`.

src/OpenDiffix.Core/Anonymizer.fs
edongashi: > the combination of ()s and >>s was here to force it to a single line

Could be. Go with whatever you think reads better.

src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -306,5 +308,6 @@ let count (executionContext: ExecutionContext) (perAidContributions: AidCountSta
     byAid
     |> Array.choose id
     |> anonymizedSum
-    |> Option.map (Math.roundAwayFromZero >> int64 >> Result.Count)
-    |> Option.defaultValue NotEnoughAIDVs
+    |> Math.roundAwayFromZero
+    |> int64
+    |> Result.Count
```
edongashi, line 312: This option stuff goes away because we do the check in `CountDistinct.Final`, right?

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -216,9 +216,12 @@ type private DiffixCountDistinct(minCount) =
       )
 
     member this.Final executionContext =
-      match Anonymizer.countDistinct executionContext aidsCount aidsPerValue with
-      | Anonymizer.Result.NotEnoughAIDVs -> Integer minCount
-      | Anonymizer.Result.Count value -> Integer(max value minCount)
+      if Option.isNone aidsCount then
```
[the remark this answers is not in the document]
pdobacz: https://github.com/diffix/reference/pull/311#discussion_r792676314

src/OpenDiffix.Core/Anonymizer.fs
pdobacz: yes, thanks to this check (and the existing one in `DiffixCount.Final`), we can make the assumption that it's a non-empty list. With that assumption `anonymizedSum` becomes optionless. See also comment near `anonymizedSum`

src/OpenDiffix.Core/Aggregator.fs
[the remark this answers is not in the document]
edongashi: Yes, but as far as I see, there is no place where we need to make the distinction between `0` and `not-initialized`.

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -188,8 +187,8 @@ type private DiffixCountDistinct(minCount) =
           match aidsPerValue.TryGetValue(value) with
           | true, aidSets -> aidSets
           | false, _ ->
-            if aidsCount = 0 then aidsCount <- aidInstances.Length
-            let aidSets = emptySets aidsCount
+            if Option.isNone aidsCount then aidsCount <- Some aidInstances.Length
```
edongashi, line 190: Maybe assert here that `aidInstances.Length > 0`.

src/OpenDiffix.Core/Aggregator.fs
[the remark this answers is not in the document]
edongashi: Ah, it already checks that `not aidInstances.IsEmpty` in the match case.

edongashi (approved):

src/OpenDiffix.Core/Anonymizer.fs
pdobacz: could work, but then how to rename the `Count` inside? or is `CountResult.Count` OK?

edongashi: Not sure, and also not important. Maybe: `CountResult.CountResult`, or `CountResult.Ok` and require explicit qualification for the union (needs an attribute).

pdobacz: `CountResult.Ok` is nice

edongashi (approved):

src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -246,9 +246,9 @@ let isLowCount (executionContext: ExecutionContext) (aidSets: HashSet<AidHash> s
   )
   |> Seq.reduce (||)
 
-type Result =
+type CountResult =
```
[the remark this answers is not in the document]
pdobacz: :+1: I force-pushed it to not multiply commits. This works nicely

edongashi (approved):

PR #312, opened 2022-01-31
Make the outlier/topRange flattening algorithm tolerate compacting

pdobacz:
This comes out of our discussion on slack about the compacting algorithm.

Since we established, that compacting only on the `topRange` end isn't a problem, let's allow it. The benefit of this is that moving on, we don't restrict the `outlierRange` to be ~smaller~ narrower than `topRange`.

I also rearrange the code a little bit in order to be able to include some explanations in the comments, I hope this works for you. We could take this further by including the check in `aidFlattening` function in the `compact...` function, but I couldn't make up my mind which option is nicer.

### DIFF
--- a/src/OpenDiffix.Core/Anonymizer.fs
+++ b/src/OpenDiffix.Core/Anonymizer.fs
@@ -47,39 +47,47 @@ let private generateNoise salt stepName stdDev noiseLayers =
 
 // Compacts flattening intervals to fit into the total count of contributors.
 // Both intervals are reduced proportionally, with `topCount` taking priority.
-// NOTE `totalCount >= outlierCount.Lower + topCount.Lower` is assumed
-let private compactFlatteningIntervals outlierCount topCount totalCount =
-  let totalAdjustment = outlierCount.Upper + topCount.Upper - totalCount
-
-  if totalAdjustment <= 0 then
-    outlierCount, topCount // no adjustment needed
+// `None` is returned in case there's not enough AIDVs for a sensible flattening.
+// `public` just to test the low-level algorithm
+let compactFlatteningIntervals outlierCount topCount totalCount =
+  if totalCount < outlierCount.Lower + topCount.Lower then
+    None
   else
-    // NOTE: at this point we know `0 <= totalAdjustment <= outlierRange + topRange` (*)
-    //       because `totalAdjustment = outlierCount.Upper + topCount.Upper - totalCount
-    //                               <= outlierCount.Upper + topCount.Upper - outlierCount.Lower - topCount.Lower`
-    //       by the condition checked in `aidFlattening`
-    let outlierRange = outlierCount.Upper - outlierCount.Lower
-    let topRange = topCount.Upper - topCount.Lower
-    // `topAdjustment` will be half of `totalAdjustment` rounded up, so it takes priority as it should
-    let outlierAdjustment = totalAdjustment / 2
-    let topAdjustment = totalAdjustment - outlierAdjustment
+    let totalAdjustment = outlierCount.Upper + topCount.Upper - totalCount
 
-    if outlierRange >= outlierAdjustment && topRange >= topAdjustment then
-      // both ranges are compacted at same rate
-      { outlierCount with Upper = outlierCount.Upper - outlierAdjustment },
-      { topCount with Upper = topCount.Upper - topAdjustment }
-    elif outlierRange < outlierAdjustment then
-      // `outlierCount` is compacted as much as possible by `outlierRange`, `topCount` takes the surplus adjustment
-      { outlierCount with Upper = outlierCount.Lower },
-      { topCount with Upper = topCount.Upper - totalAdjustment + outlierRange }
-    elif topRange < topAdjustment then
-      // vice versa
-      { outlierCount with Upper = outlierCount.Upper - totalAdjustment + topRange },
-      { topCount with Upper = topCount.Lower }
-    else
-      // Not possible. Otherwise `outlierRange + topRange < outlierAdjustment + topAdjustment = totalAdjustment` but we
-      // knew the opposite was true in (*) above
-      failwith "Not possible"
+    let compactIntervals =
+      if totalAdjustment <= 0 then
+        outlierCount, topCount // no adjustment needed
+      else
+        // NOTE: at this point we know `0 < totalAdjustment <= outlierRange + topRange` (*)
+        //       because `totalAdjustment = outlierCount.Upper + topCount.Upper - totalCount
+        //                               <= outlierCount.Upper + topCount.Upper - outlierCount.Lower - topCount.Lower`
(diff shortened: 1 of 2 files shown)


src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -47,25 +47,39 @@ let private generateNoise salt stepName stdDev noiseLayers =
...
+      { outlierCount with Upper = outlierCount.Lower },
+      { topCount with Upper = topCount.Upper - totalAdjustment + outlierRange }
+    elif topRange < topAdjustment then
+      // vice versa
+      { outlierCount with Upper = outlierCount.Upper - totalAdjustment + topRange },
+      { topCount with Upper = topCount.Lower }
+    else
+      // Not possible. Otherwise `outlierRange + topRange < outlierAdjustment + topAdjustment = totalAdjustment` but we
+      // knew the opposite was true in (*) above
+      failwith "Not possible"
```
edongashi, line 82: Should be a better error message. Or is this 100% sure to never run?

src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -47,25 +47,39 @@ let private generateNoise salt stepName stdDev noiseLayers =
...
+      { outlierCount with Upper = outlierCount.Upper - outlierAdjustment },
+      { topCount with Upper = topCount.Upper - topAdjustment }
+    elif outlierRange < outlierAdjustment then
+      // `outlierCount` is compacted as much as possible by `outlierRange`, `topCount` takes the surplus adjustment
+      { outlierCount with Upper = outlierCount.Lower },
+      { topCount with Upper = topCount.Upper - totalAdjustment + outlierRange }
+    elif topRange < topAdjustment then
+      // vice versa
+      { outlierCount with Upper = outlierCount.Upper - totalAdjustment + topRange },
+      { topCount with Upper = topCount.Lower }
```
edongashi, line 78: I'm finding it difficult to figure out what's supposed to happen. Maybe Cristian could take a look as well.

edongashi (dismissed):

src/OpenDiffix.Core/Anonymizer.fs
cristianberneanu: In cases like this, I find it better to omit the last `if` branch and instead assert the condition is true in the `else` branch.

src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -47,25 +47,39 @@ let private generateNoise salt stepName stdDev noiseLayers =
 
 // Compacts flattening intervals to fit into the total count of contributors.
 // Both intervals are reduced proportionally, with `topCount` taking priority.
+// NOTE `totalCount >= outlierCount.Lower + topCount.Lower` is assumed
```
cristianberneanu, line 50: It is better to put stuff like this in an assert rather than in a comment.

src/OpenDiffix.Core/Anonymizer.fs
cristianberneanu: I think some unit tests would be useful to ensure correctness here.

cristianberneanu: Otherwise, looks correct to me.

pdobacz: +1 unit tests, thanks for the nudge. I was considering splitting the compaction to a tiny dedicated module, and making a low-level test for that, but not sure if not overkill.

Otherwise the unit test would be very tailor-made to enact the different paths. WDYT?

src/OpenDiffix.Core/Anonymizer.fs
pdobacz: > Should be a better error message. Or is this 100% sure to never run?

So the comments aim at "proving" that 100% won't ever run. What would be better message for a runtime assertion failure like this? Do we have a precedent?

re the if/elif/else suggestion, I'll take a look, together with the proof/assertion suggestions

src/OpenDiffix.Core/Anonymizer.fs
pdobacz: it kinda is there, in the `Not possible` branch, but agreed, we could trade it for something that is put here. I'll think about this

src/OpenDiffix.Core/Anonymizer.fs
cristianberneanu: I wouldn't make a separate module just for this. It leads to overly-fragmented code.
Marking the function `public` so it can be tested would be best in my opinion.

src/OpenDiffix.Core/Anonymizer.fs
pdobacz: > In cases like this, I find it better to omit the last if

OK, I see where the confusion came from, this is a different case i think. We have 2x2=4 mutually exclusive conditions, depending on which adjustment "fits" the range. I'll refactor so this stands out better

src/OpenDiffix.Core/Anonymizer.fs
edongashi: Can be the first statement in the function. F# supports [assert](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/assertions). We should utilize it more!

pdobacz: :+1: I'm giving this bit a revamp, but I'll `assert` if still relevant afterwards

edongashi:
Build is failing

pdobacz:
> Build is failing

oh, I swapped `failwith ...` for `assert false` in the code path that's not possible, but it has different type semantics. Pushed too hastily just before leaving. I'll revert to `failwith` then.

pdobacz:
should be fine now, I also spotted a compilation warning in the tests and fixed it

edongashi (dismissed):

src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -47,39 +47,47 @@ let private generateNoise salt stepName stdDev noiseLayers =
...
+          { outlierCount with Upper = outlierCount.Lower },
+          { topCount with Upper = topCount.Upper - totalAdjustment + outlierRange }
+        | true, false ->
+          // vice versa
+          { outlierCount with Upper = outlierCount.Upper - totalAdjustment + topRange },
+          { topCount with Upper = topCount.Lower }
+        | false, false ->
+          // Not possible. Otherwise `outlierRange + topRange < outlierAdjustment + topAdjustment = totalAdjustment` but we
+          // knew the opposite was true in (*) above
+          failwith "Not possible"
```
cristianberneanu, line 88: maybe "Invalid state" or "Internal error"?

cristianberneanu (approved):

cristianberneanu (approved):

PR #313, opened 2022-02-02
Support mapping over subqueries

edongashi:


src/OpenDiffix.Core/QueryValidator.fs
```diff
@@ -48,14 +48,7 @@ let private validateSubQuery selectQuery =
   validateSelectTarget selectQuery
```
edongashi, line 48: Notice how this becomes mutually recursive. The caller controls when (and if) that happens.

src/OpenDiffix.Core/NodeUtils.fs
```diff
@@ -18,7 +18,14 @@ type NodeFunctions =
     List.map f expressions
 
   // Query
-  static member Map(query: Query, f: SelectQuery -> SelectQuery) = f query
+  static member Map(query: SelectQuery, f: Query -> Query) =
```
pdobacz, line 21: so to ELI5 this:

`Map` should step into all `Query`-children of `SelectQuery`. Those happen to be sitting in `query.From`. So here we define a function `mapSubqueries`, which needs to know how to step into `QueryRange`. This knowledge is provided by a `Map` member a several lines below....

ugh, now, I don't follow this. Somehow this particular `Map` doesn't look/behave like the others, what am I missing here?

edongashi: Your explanation is correct.

This visits queries in range of given query (not grandchildren - subqueries of subqueries, only subqueries of THIS query). I don't know, this feels consistent with other mappings to me.

pdobacz: what doesn't feel consistent is that we "cross" between traversing by `Query` into traversing by `QueryRange`, so we "change" the type of `f`, so to speak. No other `Map` seems to be doing this. This concern is super hand wavey, I'm experimenting with something which maybe will illustrate my thinking better.

edongashi: This is a design fault in typings. We encode the join tree in the same data as the range table entries. A better design would be a flat range list and a separate join tree with indexes. Anyhow, this mapping ignores the join structure as we don't care about it.

pdobacz: here, see this version https://github.com/diffix/reference/commit/42fa91c1ad27d79b9c9212ea5cde0789ab3e6932

pdobacz: So in that commit I've split your single `Map` into 3 maps, which progress. But now I think I'm seeing the problem - my version has the problem of going into grand-children, while your stops at the "first thing that looks like `Query`, and we call it an immediate child".

edongashi: > first thing that looks like Query, and we call it an immediate child

Yes, technically subqueries of a query are "immediate" range table entries (try dumping a postgres query node with subqueries). Being inside a (join) tree is unrelated to that, it just specifies the join order between RTEs.

pdobacz (approved):
works for my case, and looks good

PR #314, opened 2022-02-02
Minor unrelated tweaks

pdobacz:
A loose collection of tweaks I gathered during past several days.

If the one about collecting aggregators is not worth it for any reason I'm missing, please let me know, we'll drop it

### DIFF
--- a/src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
@@ -157,7 +157,7 @@ type Tests(db: DBFixture) =
         Rows = [ [| String "1Water" |] ]
       }
 
-    let queryResult = runQuery "SELECT CAST(id AS text) || name AS n FROM products WHERE id = 1"
+    let queryResult = runQuery "SELECT CAST(id AS text) || name AS n FROM products WHERE id = 1 GROUP BY id, name"
     queryResult |> should equal expected
 
   [<Fact>]
--- a/src/OpenDiffix.Core/Analyzer.fs
+++ b/src/OpenDiffix.Core/Analyzer.fs
@@ -377,7 +377,6 @@ let private functionSeedMaterial =
   | FloorBy -> "floor"
   | RoundBy -> "round"
   | WidthBucket -> "width_bucket"
-  | Concat -> "concat"
   | _ -> failwith "Unsupported function used for defining buckets."
 
 let rec private collectSeedMaterials rangeColumns expression =
--- a/src/OpenDiffix.Core/QueryValidator.fs
+++ b/src/OpenDiffix.Core/QueryValidator.fs
@@ -34,7 +34,7 @@ let private allowedCountUsage query =
       match args with
       | []
       | [ ColumnReference _ ] -> ()
-      | _ -> failwith "Only count(*) and count(distinct column) are supported"
+      | _ -> failwith "Only count(*), count(column) and count(distinct column) are supported"
     | _ -> ()
   )
 
@@ -48,7 +48,7 @@ let private validateSubQuery selectQuery =
   validateSelectTarget selectQuery
   validateLimitUsage selectQuery
 
-let validateSelectTarget selectQuery = selectQuery |> visit validateSubQuery
+let private validateSelectTarget selectQuery = selectQuery |> visit validateSubQuery
 
 let private validateLimitUsage selectQuery =
   if selectQuery.Limit <> None then


src/OpenDiffix.Core/NodeUtils.fs
```diff
@@ -104,7 +104,7 @@ let inline collect func node =
 let inline unwrap node =
   callUnwrap (Unchecked.defaultof<NodeFunctions>, node)
 
-/// Collects all aggregator expressions from a node.
+/// Collects all aggregators from expressions being immediate children of `node`.
```
edongashi, line 107: "immediate children of `node`" is not quite accurate. This recurses inside expressions, so if you have the following tree: `Func.Add (Agg.Count, Agg.CountDistinct)` it will collect both of the aggs.

src/OpenDiffix.Core/NodeUtils.fs
```diff
@@ -113,11 +113,15 @@ let inline collectAggregates node =
 
   node |> collect exprAggregators
 
-/// Visits all aggregate expressions in a node.
+/// Visits all aggregate expressions being immediate children of `node`.
```
edongashi, line 116: Same as above.

src/OpenDiffix.Core/NodeUtils.fs
```diff
@@ -113,11 +113,15 @@ let inline collectAggregates node =
...
 let inline visitAggregates f node =
   let rec exprVisitor f expr =
     match expr with
     | FunctionExpr (AggregateFunction (_fn, _opts), _args) as aggregateExpression -> f aggregateExpression
     | other -> other |> visit (exprVisitor f)
 
   node |> visit (exprVisitor f)
+
+let rec visitIncludingSubQueries visitor (query: SelectQuery) =
```
edongashi, line 125: Visit always includes subqueries, maybe `visitIncludingSelf`?

cristianberneanu (commented):
I don't understand why we need to collect all aggregators in bulk from a query tree.
When (or if) we will need to support pre-processing subqueries, we will need to treat pre-processing, anonymizing, post-processing aggregators differently.
For now, I think we should make `reference` always anonymize the bottom most subqueries and worry about pre-processing sensitive data later (once we fully support all the clause in a single query).

src/OpenDiffix.Core/NodeUtils.fs
pdobacz: :+1: let's discuss there

src/OpenDiffix.Core/NodeUtils.fs
pdobacz: hm, but then `Func.Add` is an expression being the immediate child. If we have `Func.Add (Agg.Count, Func.Add (Agg.CountDistinct, 3))`, then the top most `Func.Add` is the expression being the immediate child. All aggregators from it will be collected/visited.

To me the current wording implies a "too deep" collection. But I'm new to this, so I'm OK with dropping, if the change sounds off.

src/OpenDiffix.Core/NodeUtils.fs
pdobacz: `visitRecursive`?.. But I'm not convinced either...

pdobacz:
> I don't understand why we need to collect all aggregators in bulk from a query tree. When (or if) we will need to support pre-processing subqueries, we will need to treat pre-processing, anonymizing, post-processing aggregators differently. For now, I think we should make `reference` always anonymize the bottom most subqueries and worry about pre-processing sensitive data later (once we fully support all the clause in a single query).

This might be an answer to the question asked in the PR description. I'm not sure we need this at all. The question is, should queries like those added in the tests, be valid or not, given current implementation of `reference`?

I'm guessing, this particular validation pass we're discussing here, aims to protect the aggregation hooks from some pathological queries - are we OK with letting them through?

src/OpenDiffix.Core/NodeUtils.fs
edongashi: I would move this to a private function where it's required and keep the interface of this module with only basic blocks.

src/OpenDiffix.Core/NodeUtils.fs
edongashi: If the query has
```
SELECT:
  Func.Add (
    Agg.Count,
    1
  )
```

then we collect the aggregator from `SELECT::Add::Count`, which is not exactly the immediate child expression.

src/OpenDiffix.Core/NodeUtils.fs
pdobacz: Actually, that's what I started with. I'll revert :+1:

cristianberneanu:
> The question is, should queries like those added in the tests, be valid or not, given current implementation of reference?

I don't think they should be added yet, because `reference` is not yet designed to handle these cases, so they would be testing whatever behavior (correct or not) currently exists.

src/OpenDiffix.Core/NodeUtils.fs
pdobacz: ah, ok, I was parsing this like `Count` _is_ the aggregator, and it is _from_ `SELECT::Add` being immediate child expr. Hm... any counter suggestions to rephrase the original to not sound like traversing everything `from a node`? :thinking:

edongashi: Not sure for alternatives, but on callsites it will be pretty obvious what is going on.

src/OpenDiffix.Core/NodeUtils.fs
edongashi: It takes one line to generalize validators for subqueries, something like:

```fsharp
let rec validateQuery query =
  // ... validate this query

  query |> visit validateQuery // descend to subqueries
```

Or the opposite if you want bottom-up traversal.
Maybe drop it altogether?

src/OpenDiffix.Core/NodeUtils.fs
pdobacz: To me, it is obvious (or at least clear) now, but wasn't initially. But let's undo the change :+1:

pdobacz:
After sleeping on it, I think, the added test cases are testing correct behavior being: "Post-aggregation hooks are being protected from queries containing multiple LCF aggregators. Instead of producing (possibly) nonsense results, they reject the query and produce a meaningful error". The behavior was incomplete (didn't descend to subqueries/joins), and the change fills this gap.

Is this reasoning correct?

edongashi:
The hooks also do a execution time check by calling `AggregationContext.lowCountIndex` which allows only 0 or 1 matches.

pdobacz:
> The hooks also do a execution time check by calling `AggregationContext.lowCountIndex` which allows only 0 or 1 matches.

oh, so do we need this check `validateSingleLowCount` here at all :) ?

edongashi:
We need it if we want to fail early during analysis. Otherwise we have to wait until all the buckets have been built during execution which can take a while.

pdobacz:
ok since this seems to require more discussion, I'm splitting of the `validateSingleLowCount` out to a separate branch and keeping only truly _minor_ tweaks here

edongashi (approved):

PR #319, opened 2022-02-08
Fix star expansion including junk cols

edongashi:


src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
```diff
@@ -209,6 +209,10 @@ type Tests(db: DBFixture) =
       "SELECT count(*) FROM customers_small LEFT JOIN purchases ON id = cid"
       "SELECT count(*) FROM (SELECT id, cid FROM customers_small LEFT JOIN purchases ON id = cid) x"
 
+    equivalentQueries //
+      "SELECT age FROM customers"
+      "SELECT * FROM (SELECT age FROM customers) c"
```
edongashi, line 214: This test fails anyway... you can start debugging here.

pdobacz (commented):
any possibility of a unit test here?

edongashi:
> any possibility of a unit test here?

`QueryEngine.Tests.fs` already do this but for subqueries without AIDs. We have no documented behavior of AIDs from subqueries so let's wait until later.

pdobacz (approved):

Issue #318, opened 2022-02-08
Preserve target entries in plan


edongashi:
A `TargetEntry` is a superset of an expression, so there's no reason why we shouldn't carry those over to the plan.

Prerequisite for EXPLAIN. I don't see how else we can print readable column references.

PR #321, opened 2022-02-08
Support pretty printing of plans

edongashi:
Can be dumped via `System.Diagnostics.Debug.WriteLine` while testing. Needs parser support if we want `EXPLAIN`.

Looks something like this:

```
Project $0
 -> Aggregate
     Group Keys: 
     Aggregates: DiffixCount([$1, $7])
     -> InnerJoin on Equals($7, $1)
         -> Seq Scan on purchases
         -> Seq Scan on customers_small
```

if we do #318 we can make the Vars friendlier.

### DIFF
--- a/src/OpenDiffix.Core/CommonTypes.fs
+++ b/src/OpenDiffix.Core/CommonTypes.fs
@@ -454,7 +454,7 @@ module Plan =
         + $"{propLine depth}Group Keys: {String.joinWithComma groupingLabels}"
         + $"{propLine depth}Aggregates: {String.joinWithComma aggregators}"
         + childToString childPlan
-      | Plan.Unique childPlan -> "Unique" + NEWLINE + (toString (depth + 1) childPlan)
+      | Plan.Unique childPlan -> "Unique" + childToString childPlan
       | Plan.Join (leftPlan, rightPlan, joinType, condition) ->
         $"{joinType} on {condition}" + childToString leftPlan + childToString rightPlan
       | Plan.Append (leftPlan, rightPlan) -> "Append" + childToString leftPlan + childToString rightPlan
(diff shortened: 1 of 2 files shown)


src/OpenDiffix.Core/CommonTypes.fs
```diff
@@ -376,6 +426,43 @@ module Plan =
...
+      | Plan.ProjectSet (childPlan, fn, args) ->
+        $"ProjectSet {fn}({String.joinWithComma args})" + childToString childPlan
+      | Plan.Filter (childPlan, condition) -> $"Filter {condition})" + childToString childPlan
+      | Plan.Sort (childPlan, orderings) -> $"Sort {String.joinWithComma orderings}" + childToString childPlan
+      | Plan.Aggregate (childPlan, groupingLabels, aggregators) ->
+        "Aggregate"
+        + $"{propLine depth}Group Keys: {String.joinWithComma groupingLabels}"
+        + $"{propLine depth}Aggregates: {String.joinWithComma aggregators}"
+        + childToString childPlan
+      | Plan.Unique childPlan -> "Unique" + NEWLINE + (toString (depth + 1) childPlan)
```
pdobacz, line 457: why don't use `childToString` here as well?

pdobacz (approved):
one nitpick and LGTM... unless you have an idea to move all (or at least some of) the explain/toString stuff out of `CommonTypes.fs`. I feel it is slightly out of scope of a common types module. But I'm fine as is as well.

edongashi:
> I feel it is slightly out of scope of a common types module.

Yes unfortunately. But we cannot override `ToString` from outside of the module where the type was declared in.

pdobacz (approved):

PR #324, opened 2022-02-14
Prevent constant expressions usage in bucket expressions.

cristianberneanu:
Constant expressions change the bucket seed without changing the bucket semantics or included data.

### DIFF
--- a/src/OpenDiffix.Core/Analyzer.fs
+++ b/src/OpenDiffix.Core/Analyzer.fs
@@ -384,7 +384,7 @@ let private functionSeedMaterial =
 let private collectSeedMaterials rangeColumns expression =
   match expression with
   | FunctionExpr (ScalarFunction fn, args) -> functionSeedMaterial fn :: List.map (basicSeedMaterial rangeColumns) args
-  | Constant _ -> failwith "Constant expressions cand be used for defining buckets."
+  | Constant _ -> failwith "Constant expressions can not be used for defining buckets."
   | _ -> [ basicSeedMaterial rangeColumns expression ]
 
 let rec private normalizeBucketLabelExpression expression =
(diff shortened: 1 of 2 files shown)


src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -382,23 +381,26 @@ let private functionSeedMaterial =
   | WidthBucket -> "width_bucket"
   | _ -> failwith "Unsupported function used for defining buckets."
 
-let rec private collectSeedMaterials rangeColumns expression =
+let private collectSeedMaterials rangeColumns expression =
   match expression with
-  | FunctionExpr (ScalarFunction Cast, [ expression; _type ]) -> collectSeedMaterials rangeColumns expression
   | FunctionExpr (ScalarFunction fn, args) -> functionSeedMaterial fn :: List.map (basicSeedMaterial rangeColumns) args
+  | Constant _ -> failwith "Constant expressions cand be used for defining buckets."
```
pdobacz, line 387: typo `can't`

also, I've realized now, we're doing these "by-the-way" validations for non-anonymizing queries as well. I'm working on related code now, I'll see what can be done about this.

To illustrate my point:

```
dotnet run --project reference/src/OpenDiffix.CLI/ -- -f reference/data/data.sqlite -q "SELECT *, age + 5 FROM customers;"
ERROR: Unsupported function used for defining buckets.
```

pdobacz (approved):

src/OpenDiffix.Core/Analyzer.fs
cristianberneanu: Yeah, that is a problem, good point. We probably need to create the SQL seed conditionally, only when an anonymizing aggregate is used (so that it works for Desktop statements also).

pdobacz: In the end, I decided to propose my PR #325 without any attempts to tackle this. Maybe let's work towards merging our currently open PRs and double-down over this problem in next step.

PR #325, opened 2022-02-14
Block anonymizing queries with subqueries/WHERE/JOIN

### LINKED ISSUE #323: Block anonymizing queries with subqueries/WHERE/JOIN
In order to conform to Elm diffix.

We support a simple choice of allowed queries:
1/ there are subqueries/joins/wheres present anywhere in the query - no part of the query can touch sensitive (aka personal) data
2/ there aren't subqueries/joins/wheres and we touch sensitive data - do anonymization according to Elm

pdobacz:
Closes #323.

As title + I spotted a bug related to joins. If the second (right) joined table had it's _first_ column in the target list, it would end up being treated as the first (left) table, resulting in an out of bounds error, e.g.:

```dotnet run --project reference/src/OpenDiffix.CLI/ -- -f reference/data/data.sqlite -q "SELECT b.age FROM customers as a JOIN customers as b ON a.id=b.id"```

```ERROR: The index was outside the range of elements in the list. (Parameter 'n')```

### DIFF
--- a/src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
@@ -349,7 +349,7 @@ type Tests(db: DBFixture) =
   [<Fact>]
   let ``Detect queries joining tables with AID columns`` () =
     ensureQueryFails
-      "SELECT 1 FROM products JOIN customers_small ON true"
+      "SELECT price FROM products JOIN customers_small ON true"
       "JOIN in anonymizing queries is not currently supported"
 
   [<Fact>]
--- a/src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
@@ -338,94 +338,25 @@ type Tests(db: DBFixture) =
     result.Having |> should equal expected
 
   [<Fact>]
-  let ``Analyze JOINs`` () =
-    let result = analyzeQuery "SELECT count(*) FROM customers_small JOIN purchases ON id = purchases.cid"
-
-    let condition =
-      FunctionExpr(ScalarFunction Equals, [ ColumnReference(4, IntegerType); ColumnReference(7, IntegerType) ])
-
-    let customers = getTable "customers_small"
-    let purchases = getTable "purchases"
-
-    let expectedFrom =
-      Join
-        {
-          Type = JoinType.InnerJoin
-          Left = RangeTable(customers, customers.Name)
-          Right = RangeTable(purchases, purchases.Name)
-          On = condition
-        }
-
-    result.From |> should equal expectedFrom
-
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/OpenDiffix.Core/Planner.fs [hunk 1]
```diff
@@ -46,7 +46,7 @@ let rec private collectSetFunctions expression =
 
 let private planJoin join columnIndices =
   let leftColumnCount = QueryRange.columnsCount join.Left
-  let leftIndices, rightIndices = List.partition ((>=) leftColumnCount) columnIndices
+  let leftIndices, rightIndices = List.partition ((>) leftColumnCount) columnIndices
```
pdobacz, line 49: bugfix

src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
```diff
@@ -337,99 +337,19 @@ type Tests(db: DBFixture) =
 
     result.Having |> should equal expected
 
+  // NOTE: We do QueryValidator testing in its respective test module. Here we just check, if it's invoked at all based
```
pdobacz, line 340: I'm not super happy about this part, I was toying with the idea of splitting the `is-it-anonymizing` logic part out somewhere and test it properly. But then I realized, that we're doing some validations while analyzing, and there we don't discriminate on anonymizing/non-anonymizing, so possibly this requires a larger revamp.

For now keeping simple and leaving as is, with this comment.

src/OpenDiffix.Core/Planner.fs
edongashi: `(>) leftColumnCount` is very confusing because it actually means `fun x -> x < leftColumnCount`. Can we please expand to full form?

edongashi (dismissed):

src/OpenDiffix.Core/Planner.fs
pdobacz: You're right! I also got confused by that, because the docs for `partition` say it returns: `A list containing the elements for which the predicate evaluated to false and a list containing the elements for which the predicate evaluated to true.`.

But it actually is the other way around! 

The description has it right `two collections, containing the elements for which the given predicate returns True and False respectively` :exploding_head:

pdobacz:
build fails with current master merged in... hold on

pdobacz:
should be good now

src/OpenDiffix.Core/Planner.fs
pdobacz: it's fixed in the fresh {version of docs](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-listmodule.html)

edongashi: `[1; 2; 3] |> List.map((-) 1)` results in `[0; -1; -2]` because the args are the other way around!

edongashi (approved):

PR #327, opened 2022-02-15
Fix implicit buckets for constants + SQL noise layers only on anonymization

pdobacz:
Two seemingly unrelated fixes, but the implementation approach taken in the [Fix implicit buckets for constants by filtering out](https://github.com/diffix/reference/pull/327/commits/a62b15ddd937ce53179772a9750a381373e3dc65) relies a bit on the other one.

Commits:

1. First commit causes SQL noise computation to be done only when it is really needed, related to [this thread in an earlier PR](https://github.com/diffix/reference/pull/324#discussion_r805977680). This fixes the non-anonymizing `SELECT age + 5 FROM customers`. We have swapped the order of `rewriteQuery` and `computeNoiseLayers`. A few notes on this in particular:
    1. In this order, the anonymizing aggregates are more readily accessible for the condition we're imposing
    1. This order is the same as in `pg_diffix`, so I figured there's nothing fundamentally wrong here
    2. I went through the rewriting code and didn't find anything that would have broken things, but I might have missed something
    3. It kinda makes more sense conceptually to rewrite first and then compute noise layers, but not sure here either
2. Second commit fixes the anonymizing `SELECT 1 FROM customers;` and the like.
3. Third commit is a patch for the second, as suggested by Edon. In the case like that, we shouldn't apply the LCF, as there the information released is equivalent to `SELECT count(*) FROM customers;`. I decided to refactor the nested `if`s into a `match`, laying out the 4 cases more explicitly

### DIFF
--- a/src/OpenDiffix.Core/Analyzer.fs
+++ b/src/OpenDiffix.Core/Analyzer.fs
@@ -323,9 +323,14 @@ let private addLowCountFilter aidColumnsExpression selectQuery =
   let doesAggregation = selectedExpressions |> List.forall Expression.isScalar |> not
   let onlyConstantsSelected = List.isEmpty nonConstantExpressions
 
-  match (doesGrouping, doesAggregation, onlyConstantsSelected) with
-  | (false, false, false) ->
+  if not doesGrouping && not doesAggregation then
     // Non-grouping, non-aggregate query; group implicitly and expand
+    let having =
+      if not onlyConstantsSelected then
+        Expression.makeAnd lowCountFilter selectQuery.Having
+      else
+        // All selected expressions are constants; no low-count filter in line with global anonymized count.
+        selectQuery.Having
 
     { selectQuery with
         TargetList =
@@ -337,23 +342,12 @@ let private addLowCountFilter aidColumnsExpression selectQuery =
           :: selectQuery.TargetList
         // No need to group implicitly by constants, which are invalid bucket definitions anyway.
         GroupBy = nonConstantExpressions
-        Having = Expression.makeAnd lowCountFilter selectQuery.Having
+        Having = having
     }
-  | (false, false, true) ->
-    // All selected expressions are constants, we expand for a result in line with global anonymized count.
-    { selectQuery with
-        TargetList =
-          {
-            Expression = bucketExpand aidColumnsExpression
-            Alias = ""
-            Tag = JunkTargetEntry
-          }
-          :: selectQuery.TargetList
-    }
-  | (false, true, _) ->
+  else if not doesGrouping && doesAggregation then
     // Non-grouping aggregate query; do nothing.
     selectQuery
-  | (true, _, _) ->
+  else
     // Grouping query; add LCF to HAVING
     { selectQuery with
         Having = Expression.makeAnd lowCountFilter selectQuery.Having
--- a/src/OpenDiffix.Core/Analyzer.fs
+++ b/src/OpenDiffix.Core/Analyzer.fs
@@ -315,6 +315,9 @@ let private addLowCountFilter aidColumnsExpression selectQuery =
       |> List.map (fun selectedColumn -> selectedColumn.Expression)
 
     if selectedExpressions |> List.forall Expression.isScalar then
+      // No need to group implicitly by constants, which are invalid bucket definitions anyway.
+      let implicitGroupExpressions = selectedExpressions |> List.filter (Expression.isConstant >> not)
+
       // Non-grouping & non-aggregating query; group implicitly and expand
       let bucketCount = Expression.makeAggregate DiffixCount [ aidColumnsExpression ]
       let bucketExpand = Expression.makeSetFunction GenerateSeries [ bucketCount ]
@@ -323,7 +326,7 @@ let private addLowCountFilter aidColumnsExpression selectQuery =
           TargetList =
             { Expression = bucketExpand; Alias = ""; Tag = JunkTargetEntry }
             :: selectQuery.TargetList
-          GroupBy = selectedExpressions
+          GroupBy = implicitGroupExpressions
           Having = Expression.makeAnd lowCountFilter selectQuery.Having
       }
     else
@@ -352,14 +355,6 @@ let private rewriteQuery anonParams (selectQuery: SelectQuery) =
 // ----------------------------------------------------------------
 // Noise layers
 // ----------------------------------------------------------------
-
-let private collectGroupingExpressions selectQuery : Expression list =
-  if List.isEmpty selectQuery.GroupBy then
-    selectQuery.TargetList
-    |> List.map (fun selectedColumn -> selectedColumn.Expression)
-    |> List.filter Expression.isScalar
-  else
-    selectQuery.GroupBy
 
 let private basicSeedMaterial rangeColumns expression =
   match expression with
@@ -407,8 +402,7 @@ let private computeNoiseLayers anonParams query =
   let rangeColumns = collectRangeColumns anonParams query.From
 
   let sqlSeed =
-    query
-    |> collectGroupingExpressions
+    query.GroupBy
     |> Seq.map (
       normalizeBucketLabelExpression
       >> collectSeedMaterials rangeColumns
--- a/src/OpenDiffix.Core/Analyzer.fs
+++ b/src/OpenDiffix.Core/Analyzer.fs
@@ -418,6 +418,18 @@ let private computeNoiseLayers anonParams query =
 
   { BucketSeed = sqlSeed }
 
+// NOTE: We do not check subqueries, as they aren't supported in conjunction with anonymization.
+let private hasAnonymizingAggregates query =
+  query
+  |> collectAggregates
+  |> List.filter (
+    function
+    | FunctionExpr (AggregateFunction (fn, opts), _) -> Aggregator.isAnonymizing (fn, opts)
+    | _ -> false
+  )
+  |> List.isEmpty
+  |> not
+
 // ----------------------------------------------------------------
 // Public API
 // ----------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


pdobacz:
whoops, I just spotted an error - in the "new case" I describe in 3. in the description, we shouldn't apply the LCF, but we **should** anonymize the count... let me fix this

pdobacz:
OK, I've force pushed the fixed Commit 3. It still fails `prop-tests` against `pg_diffix`, as I think `pg_diiffix` applies the LCF in the "new case". I'll handle this in the next step.

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -304,35 +304,56 @@ let private rewriteToDiffixAggregate aidColumnsExpression query =
...
-      // Non-grouping aggregate query; do nothing
-      selectQuery
-  else
+  let selectedExpressions =
+    selectQuery.TargetList
+    |> List.map (fun selectedColumn -> selectedColumn.Expression)
+
+  let nonConstantExpressions = selectedExpressions |> List.filter (Expression.isConstant >> not)
+
+  let isGrouping = List.isEmpty selectQuery.GroupBy |> not
```
cristianberneanu, line 322: `isGrouping` -> `hasGroups/doesGrouping`?

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -304,35 +304,56 @@ let private rewriteToDiffixAggregate aidColumnsExpression query =
...
+  let isAggregate = selectedExpressions |> List.forall Expression.isScalar |> not
```
cristianberneanu, line 323: `isAggregate` -> `hasAggregators/doesAggregation`?

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -304,35 +304,56 @@ let private rewriteToDiffixAggregate aidColumnsExpression query =
...
+        TargetList =
+          {
+            Expression = bucketExpand aidColumnsExpression
+            Alias = ""
+            Tag = JunkTargetEntry
+          }
+          :: selectQuery.TargetList
+        // No need to group implicitly by constants, which are invalid bucket definitions anyway.
+        GroupBy = nonConstantExpressions
+        Having = Expression.makeAnd lowCountFilter selectQuery.Having
```
cristianberneanu, line 340: I think it would be simpler to add the LCF conditionally here and drop the following case.

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -304,35 +304,56 @@ let private rewriteToDiffixAggregate aidColumnsExpression query =
...
+    selectQuery.TargetList
+    |> List.map (fun selectedColumn -> selectedColumn.Expression)
+
+  let nonConstantExpressions = selectedExpressions |> List.filter (Expression.isConstant >> not)
+
+  let isGrouping = List.isEmpty selectQuery.GroupBy |> not
+  let isAggregate = selectedExpressions |> List.forall Expression.isScalar |> not
+  let onlyConstantsSelected = List.isEmpty nonConstantExpressions
+
+  match (isGrouping, isAggregate, onlyConstantsSelected) with
```
cristianberneanu, line 326: This is so much harder to parse now...

cristianberneanu (commented):
What about these cases?
```SQL
SELECT 1, COUNT(*) FROM table
SELECT 1, COUNT(*) FROM table GROUP BY 1
SELECT 1, col GROUP BY 1, 2
```

src/OpenDiffix.Core/Analyzer.fs
pdobacz: harder than which version? Note, that this commit adds a 3rd layer of nested if. I started off with nested ifs and it was much worse off. So comparing against master isn't fair.

Also current refactored form comes from our convo with Edon, where we settled on this being a 4th top-level path which should be handled, so the `match with` version captures it - 4 separate top-level paths.

I needed to translate this into the `pg_diffix` implementation, I can't imagine doing this without the explicit, 4-legged match.

src/OpenDiffix.Core/Analyzer.fs
pdobacz: let's take it to the other thread

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -315,6 +315,9 @@ let private addLowCountFilter aidColumnsExpression selectQuery =
       |> List.map (fun selectedColumn -> selectedColumn.Expression)
 
     if selectedExpressions |> List.forall Expression.isScalar then
+      // No need to group implicitly by constants, which are invalid bucket definitions anyway.
+      let implicitGroupExpressions = selectedExpressions |> List.filter (Expression.isConstant >> not)
```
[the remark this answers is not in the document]
pdobacz: Hm, if I do that, it fails one of the tests. I'll investigate why.

src/OpenDiffix.Core/Analyzer.fs
cristianberneanu: The ifs don't all have to be nested, most could be on the same level. What do you think about this:
```F#
if not isGrouping && not isAggregate then // no aggregation
  let havingClause =
    if onlyConstantsSelected then
      Expression.makeAnd lowCountFilter selectQuery.Having
    else
      query.Having
  ...
else if not isGrouping && isAggregate then // global aggregation
   ...
else // grouped aggregation
  ...
```

cristianberneanu: My main problem with the current version is that I have to scroll up at the beginning of the match when trying to understand each branch.

pdobacz: This could work, I thought you were referring to the entire combined refactor. Let's do the `else if` version, however I still prefer the flatness of the `match`, but that's just personal preference.

I've also seen an approach with boolean aliases like `let GROUPING, NON_GROUPING = true, false`, I'm still not sure if I like it though.

pdobacz: Nested if is done in 654eb5b7, let me know if you like it better than a `match with` with the boolean aliases for example

cristianberneanu: Yes, I think it looks better now.

src/OpenDiffix.Core/Analyzer.fs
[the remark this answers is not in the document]
pdobacz: the test failed, because it compared SQL layer seed and https://github.com/diffix/reference/blob/master/src/OpenDiffix.Core.Tests/Analyzer.Tests.fs#L306-L311 didn't have `Normalizer.normalize` in the pipeline. But I think it should, why would we want it to be the odd one out?

If I put `Normalizer.normalize` in, there's another test failing, one about consistency of seeding numeric ranges. If the `normalize` should indeed be there like in production code, then this indicates we have a bug, I'll investigate

pdobacz:
> What about these cases?
> 
> ```sql
> SELECT 1, COUNT(*) FROM table
> SELECT 1, COUNT(*) FROM table GROUP BY 1
> SELECT 1, col GROUP BY 1, 2
> ```

Amidst the other discussions, I got the result:

```
1,206
Constant expressions can not be used for defining buckets.
Constant expressions can not be used for defining buckets.
```

respectively. I think legit. Should we have that in tests?

cristianberneanu:
But they should all be allowed right? My point is that we need to remove constant expressions from the grouping clause (explicit or implicit). I think that should handle everything more elegantly.

pdobacz:
> But they should all be allowed right? My point is that we need to remove constant expressions from the grouping clause (explicit or implicit). I think that should handle everything more elegantly.

I think only the first should be allowed, the `1` there is irrelevant, it's just a global aggregation query with a useless `1`. The other two should not be allowed, as they define buckets using constants.

> need to remove constant expressions from the grouping clause (explicit or implicit).

This is what we're doing now. We rewrite first and then look for constants in the grouping clause. The latter step doesn't care about whether the grouping has been explicit or implicit.

And the results I posted confirm that the implementation follows this.

Are we good here? ... Oh wait, I see your point. you want to remove the constants, instead of rejecting queries that end up with  them in the group clause? That's possible, but maybe could be an independent follow-up? This PR is getting big, with a lot of bumps.

src/OpenDiffix.Core/Analyzer.fs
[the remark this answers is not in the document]
pdobacz: For future reference: After some discussion elsewhere, we decided to just comment out the offending test cases, and fix in a follow-up PR. The fix I proposed which didn't get into `master` is in 29da4c72ef7.

cristianberneanu:
> I think only the first should be allowed, the 1 there is irrelevant, it's just a global aggregation query with a useless 1. The other two should not be allowed, as they define buckets using constants.

In the first query, the `1` is useless as well. And it also defines a bucket with a constant. In my mind, they all seem similar, and we should either handle all of them or none. But it is fine to leave if for another PR.

src/OpenDiffix.Core/Analyzer.fs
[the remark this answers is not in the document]
pdobacz: And the fix for the original request in this thread is 593f3e2d, LMK if this is what you've meant

cristianberneanu (approved):

pdobacz:
> > I think only the first should be allowed, the 1 there is irrelevant, it's just a global aggregation query with a useless 1. The other two should not be allowed, as they define buckets using constants.
> 
> In the first query, the `1` is useless as well. And it also defines a bucket with a constant. In my mind, they all seem similar, and we should either handle all of them or none. But it is fine to leave if for another PR.

One remark, so that we're clear: "it also defines a bucket with a constant" - it doesn't. It falls into the `if not doesGrouping && doesAggregation` category and remains unchanged.

cristianberneanu:
> : "it also defines a bucket with a constant" - it doesn't.

You're right, I confused it with the `SELECT 1 FROM table` case.
What I am trying to say is that if we support `SELECT 1 FROM table`, we should also support `SELECT 1 FROM table GROUP BY 1`. Similarly for `SELECT 1, col FROM table` and `SELECT 1, col FROM table GROUP BY 1, 2`.

PR #328, opened 2022-02-22
Add Mode anonymization parameter and implement Untrusted Mode

### LINKED ISSUE #316: Support trusted/untrusted analyst modes.
Untrusted mode should apply additional restrictions to query handling (as per the Elm spec).

pdobacz:
Closes #316

### DIFF
--- a/src/OpenDiffix.CLI/JsonEncodersDecoders.fs
+++ b/src/OpenDiffix.CLI/JsonEncodersDecoders.fs
@@ -48,6 +48,7 @@ module private rec Encoders =
           (ap.TableSettings
            |> Map.toList
            |> List.map (fun (table, settings) -> {| Table = table; Settings = settings |}))
+        AccessLevel = ap.AccessLevel
         LowThreshold = ap.Suppression.LowThreshold
         LowMeanGap = ap.Suppression.LowMeanGap
         LowLayerSd = ap.Suppression.LayerSD
@@ -73,6 +74,27 @@ module private rec Encoders =
       let response = {| Success = false; error = errorMsg |}
       autoEncode response
 
+let private anonymizationModeDecoder: Decoder<AccessLevel> =
+  Decode.string
+  |> Decode.andThen (
+    function
+    | "publish_trusted" -> Decode.succeed PublishTrusted
+    | "publish_untrusted" -> Decode.succeed PublishUntrusted
+    | "direct" -> Decode.succeed Direct
+    | invalid -> Decode.fail (sprintf "Failed to decode `%s`, not a valid `AnonymizationMode`" invalid)
+  )
+
+let private anonymizationModeEncoder (point: AccessLevel) =
+  match point with
+  | PublishTrusted -> "publish_trusted"
+  | PublishUntrusted -> "publish_untrusted"
+  | Direct -> "direct"
+  |> Encode.string
+
+let private anonymizationModeExtra =
+  Extra.empty
+  |> Extra.withCustom anonymizationModeEncoder anonymizationModeDecoder
+
 // ----------------------------------------------------------------
 // Public API
 // ----------------------------------------------------------------
@@ -93,4 +115,4 @@ let encodeBatchRunResult (time: System.DateTime) version (queryResults: QueryRes
     .ToString()
 
 let decodeRequestParams content =
-  Decode.Auto.fromString<QueryRequest list> (content, caseStrategy = SnakeCase)
+  Decode.Auto.fromString<QueryRequest list> (content, caseStrategy = SnakeCase, extra = anonymizationModeExtra)
--- a/src/OpenDiffix.CLI/JsonEncodersDecoders.fs
+++ b/src/OpenDiffix.CLI/JsonEncodersDecoders.fs
@@ -81,11 +81,11 @@ let private anonymizationModeDecoder: Decoder<AccessLevel> =
     | "publish_trusted" -> Decode.succeed PublishTrusted
     | "publish_untrusted" -> Decode.succeed PublishUntrusted
     | "direct" -> Decode.succeed Direct
-    | invalid -> Decode.fail (sprintf "Failed to decode `%s`, not a valid `AnonymizationMode`" invalid)
+    | invalid -> Decode.fail $"Failed to decode {invalid}, not a valid `AccessLevel`"
   )
 
-let private anonymizationModeEncoder (point: AccessLevel) =
-  match point with
+let private anonymizationModeEncoder (accessLevel: AccessLevel) =
+  match accessLevel with
   | PublishTrusted -> "publish_trusted"
   | PublishUntrusted -> "publish_untrusted"
   | Direct -> "direct"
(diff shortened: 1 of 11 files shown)


src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -419,13 +424,30 @@ let rec private normalizeBucketLabelExpression expression =
     FunctionExpr(ScalarFunction fn, List.map normalizeBucketLabelExpression args @ extraArgs)
   | _ -> expression
 
+let private untrustedAllowsRange arg =
+  match arg with
+  // "money-style" numbers, i.e. 1, 2, or 5 preceeded by or followed by zeros: ⟨... 0.1, 0.2, 0.5, 1, 2, 5, 10, 20, ...⟩
+  | Constant (Real c) -> Regex("^(1|2|5)\.0+e(\-|\+)[0-9]+$").Match(sprintf "%.15e" c).Success
```
pdobacz, line 430: I considered a more "mathy" formulation of this condition, but deciding this through a string representation seemed more promising in terms of clarity.

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -419,13 +424,30 @@ let rec private normalizeBucketLabelExpression expression =
...
+  expression
+
 let private computeNoiseLayers anonParams query =
   let rangeColumns = collectRangeColumns anonParams query.From
 
   let sqlSeed =
     query.GroupBy
     |> Seq.map (
       normalizeBucketLabelExpression
+      >> validateBucketLabelExpression anonParams.AccessLevel
```
pdobacz, line 450: Ideally, this should be under `QueryValidator`, just like other validations that the seed calculation is currently doing. I'd like to proceed as is and consider a larger refactoring in a future PR.

src/OpenDiffix.CLI/JsonEncodersDecoders.fs
```diff
@@ -73,6 +74,27 @@ module private rec Encoders =
...
       autoEncode response
 
+let private anonymizationModeDecoder: Decoder<AccessLevel> =
+  Decode.string
+  |> Decode.andThen (
+    function
+    | "publish_trusted" -> Decode.succeed PublishTrusted
+    | "publish_untrusted" -> Decode.succeed PublishUntrusted
+    | "direct" -> Decode.succeed Direct
+    | invalid -> Decode.fail (sprintf "Failed to decode `%s`, not a valid `AnonymizationMode`" invalid)
```
edongashi, line 84:
```suggestion
    | invalid -> Decode.fail $"Failed to decode `{invalid}`, not a valid `AnonymizationMode`"
```

src/OpenDiffix.CLI/Program.fs
```diff
@@ -39,6 +40,8 @@ type CliArguments =
         + "The db_path inside should be relative to this file's path."
       | Query_Stdin -> "Reads the query from standard in."
       | Salt _ -> "The salt value to use when anonymizing the data. Changing the salt will change the result."
+      | Mode _ ->
+        "Anonymization mode: 'untrusted' mode is more restrictive and assumes secure anonymization despite the analyst actively attempting to de-anonymize the results."
```
edongashi, line 44: Exceeds line length, maybe split or shorten. Something like `'untrusted' assumes a malicious analyst actively attempting to de-anonymize the results`

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -1,5 +1,7 @@
 module rec OpenDiffix.Core.Analyzer
 
+open System.Text.RegularExpressions
+
 open AnalyzerTypes
 open NodeUtils
```
edongashi, line 6: AFAIK we don't separate stdlib like in pg_diffix here. We start with `System.*` then others.

src/OpenDiffix.Core/Analyzer.fs
edongashi: Interesting! I'm not sure how to judge this, I guess `*.5` plays well with binary representation.

Some notes about regex use:

- Pattern should be cached outside of the function, or use `RegExp.IsMatch(string, pattern)`.
- Use `IsMatch` instead of `Match().Success`
- Groups don't need to capture since you only check for match success
- `(1|2|5)` can be simplified to `[125]`
- `[0-9]` can be simplified to `\d`
- `(\-|\+)` can be simplified to `[-+]`

src/OpenDiffix.Core/CommonTypes.fs
```diff
@@ -140,11 +140,17 @@ type SuppressionParams =
   }
   static member Default = { LowThreshold = 2; LayerSD = 1.; LowMeanGap = 2. }
 
+type AccessLevel =
```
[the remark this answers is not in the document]
edongashi: I see this called as `anonymizationMode` or just `mode` in other places, and here `AccessLevel`. We should ensure consistency from the get-go.

src/OpenDiffix.CLI/JsonEncodersDecoders.fs
```diff
@@ -73,6 +74,27 @@ module private rec Encoders =
...
+  )
+
+let private anonymizationModeEncoder (point: AccessLevel) =
```
edongashi, line 87: `point`?

src/OpenDiffix.CLI/Program.fs
```diff
@@ -92,6 +95,12 @@ let toSalt =
   | Some (salt: string) -> Text.Encoding.UTF8.GetBytes(salt)
   | _ -> [||]
 
+let toMode =
+  function
+  | Some ("publish_unstrusted") -> PublishUntrusted
+  | Some ("publish_direct") -> Direct
```
edongashi, line 101: Redundant parens.

src/OpenDiffix.Core/CommonTypes.fs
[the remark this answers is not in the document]
pdobacz: ouch, grep-fail, my bad, sorry

src/OpenDiffix.Core/Analyzer.fs
pdobacz: > I guess *.5 plays well with binary representation.

yeah, the others are a bit funky, hence the `%.15e`. I'm not 100% it's completely air tight, but should be good enough, even if some edge cases :thinking:.

The alternative would be to disassemble the floating point binary representation and dig into that. The regex version has the advantage that it reads "similar" to the definition from the specs - "if it looks like money-style number, it is a money-style number"

> Some notes about regex use

Thx!

src/OpenDiffix.CLI/Program.fs
pdobacz: yup, there's a nice and short definition in the spec, I'll use this

edongashi (approved):

pdobacz:
Apologies about the rough edges. I've also found one thing in the CLI part. This should be better now.

src/OpenDiffix.Core/Analyzer.fs
edongashi: Do you intend to push any change for this?

edongashi (dismissed):

src/OpenDiffix.Core/Analyzer.fs
pdobacz: missed this one, crazy day today :bow: And I think gh sometimes shows it for me and sometimes doesn't...

edongashi (approved):

PR #329, opened 2022-02-28
Additional test: ceil and round blocked in untrusted mode + rename

pdobacz:
This is just a small fixup of #328. This adds a test that `ceil` and `round` behave as expected in untrusted analyst mode. Also smuggles a rename of a function, whose name I found very confusing on second visit.

EDIT: after discussion, it turned out that we actually want to admit `ceil` and `round`. While doing so, I found that non-generalized queries were invalidly blocked, so fixed that as well.

### DIFF
--- a/src/OpenDiffix.Core/Analyzer.fs
+++ b/src/OpenDiffix.Core/Analyzer.fs
@@ -430,11 +430,16 @@ let private isMoneyStyle arg =
   | Constant (Integer c) -> Regex.IsMatch($"%i{c}", "^[125]0*$")
   | _ -> false
 
-let private validateBucketLabelExpression accessLevel expression =
+let private validateGeneralization accessLevel expression =
   if accessLevel = PublishUntrusted then
     match expression with
-    | FunctionExpr (ScalarFunction FloorBy, [ _; arg ]) when isMoneyStyle arg -> ()
+    | FunctionExpr (ScalarFunction fn, [ _ ]) when List.contains fn [ Floor; Ceil; Round ] -> ()
+    | FunctionExpr (ScalarFunction fn, [ _; arg ]) when
+      List.contains fn [ FloorBy; CeilBy; RoundBy ] && isMoneyStyle arg
+      ->
+      ()
     | FunctionExpr (ScalarFunction Substring, [ _; fromArg; _ ]) when fromArg = (1L |> Integer |> Constant) -> ()
+    | ColumnReference _ -> ()
     | _ -> failwith "Generalization used in the query is not allowed in untrusted access level"
 
   expression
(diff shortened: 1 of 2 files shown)


edongashi (dismissed):

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -433,7 +433,7 @@ let private untrustedAllowsRange arg =
 let private validateBucketLabelExpression accessLevel expression =
   if accessLevel = PublishUntrusted then
     match expression with
-    | FunctionExpr (ScalarFunction FloorBy, [ _; arg ]) when untrustedAllowsRange arg -> ()
+    | FunctionExpr (ScalarFunction FloorBy, [ _; arg ]) when isMoneyStyle arg -> ()
```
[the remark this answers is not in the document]
pdobacz: Elm spec specifically allows floor_by only. I haven't given this too much thought though, it seems, there is no loss of anonymity if we allow them. Let me ask to be sure, before I merge this.

pdobacz:
@edongashi it turned out that this change should actually take the opposite direction: we actually want to admit ceil and round. While doing so, I found that non-generalized queries were invalidly blocked, so fixed that as well.

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -423,18 +423,23 @@ let rec private normalizeBucketLabelExpression expression =
...
-let private untrustedAllowsRange arg =
+let private isMoneyStyle arg =
   match arg with
   // "money-style" numbers, i.e. 1, 2, or 5 preceeded by or followed by zeros: ⟨... 0.1, 0.2, 0.5, 1, 2, 5, 10, 20, ...⟩
   | Constant (Real c) -> Regex.IsMatch($"%.15e{c}", "^[125]\.0+e[-+]\d+$")
   | Constant (Integer c) -> Regex.IsMatch($"%i{c}", "^[125]0*$")
   | _ -> false
 
-let private validateBucketLabelExpression accessLevel expression =
+let private validateGeneralization accessLevel expression =
```
edongashi, line 433: I think `validateBucketLabelExpression` is the more general term.

edongashi (approved):

src/OpenDiffix.Core/Analyzer.fs
pdobacz: So I thought initially, but then I observed this is actually that specific. Note that we also do other kinds of bucket label expression validations. Once untrusted analyst mode concerns anything else than generalizations, we'll rename/refactor as needed. OTOH no-one wiill get the wrong idea that this functions _is the_ bucket label expression validation routine.

PR #330, opened 2022-03-02
Validate bucket functions usage in non-direct mode.

cristianberneanu:


src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -430,7 +430,25 @@ let private isMoneyStyle arg =
...
+  | ColumnReference _ -> true
+  | _ -> false
+
+let private isConstant arg =
+  match arg with
+  | Constant _ -> true
+  | _ -> false
+
 let private validateGeneralization accessLevel expression =
+  if accessLevel <> Direct then
```
pdobacz, line 444: Just FYI: after the initial merge, I had a plan to refactor these bits and move the validation to `QueryValidator`, I'll move this new bit over too (I don't remember exactly why was this troublesome at first).

src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
```diff
@@ -415,6 +415,16 @@ type Tests(db: DBFixture) =
     analyzeQueryUntrusted "SELECT floor(age) from customers" |> ignore
     analyzeQueryUntrusted "SELECT age from customers" |> ignore
 
+  [<Fact>]
+  let ``Detect queries with disallowed bucket functions calls`` () =
+    ensureQueryFailsUntrusted
```
pdobacz, line 420: is `...Untrusted` necessary here? this should fail in trusted mode as well.

src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
```diff
@@ -415,6 +415,16 @@ type Tests(db: DBFixture) =
...
+      "SELECT round(2, age) from customers"
```
pdobacz, line 421: let's make the second arg here valid, i.e. `2` or `2.0`, or make that the third tested case.

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -430,7 +430,25 @@ let private isMoneyStyle arg =
...
+    match expression with
+    | FunctionExpr (ScalarFunction _, primaryArg :: _) when not (isColumnReference primaryArg) ->
+      failwith "Primary argument for a bucket function has to be a simple column reference."
+    | FunctionExpr (ScalarFunction _, _ :: secondaryArgs) when List.exists (isConstant >> not) secondaryArgs ->
+      failwith "Secondary arguments for a bucket function have to be a constants."
```
pdobacz, line 449: `have to be constants``

src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
cristianberneanu: If both arguments are constants, then the function call is replaced by a constant during normalization.

src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
pdobacz: exactly, this is why we should `ensureQueryFails` instead, otherwise you're implying the validation is maybe done in untrusted mode only. `trusted` mode is "enough" for the query to fail.

src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
pdobacz: aaah, right! that's good then. I'm not sure whether and how to test it then, up to you. In a way, it is good that it doesn't fall into some "strange" path and produce an irrelevant error message.

PR #331, opened 2022-03-03
Refactor query compilation.

cristianberneanu:
Simplifies and improves compilation of anonymizing queries.

### DIFF
--- a/src/OpenDiffix.Core.Tests/QueryValidator.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/QueryValidator.Tests.fs
@@ -2,33 +2,50 @@ module OpenDiffix.Core.QueryValidatorTests
 
 open Xunit
 
-let testTable: Table =
-  {
-    Name = "table"
-    Columns =
-      [
-        { Name = "str_col"; Type = StringType }
-        { Name = "int_col"; Type = IntegerType } // AID column
-        { Name = "float_col"; Type = RealType }
-        { Name = "bool_col"; Type = BooleanType }
-      ]
+let testTables =
+  [
+    {
+      Name = "anon_table"
+      Columns =
+        [
+          { Name = "id"; Type = IntegerType }
+          { Name = "str_col"; Type = StringType }
+          { Name = "int_col"; Type = IntegerType }
+          { Name = "float_col"; Type = RealType }
+          { Name = "bool_col"; Type = BooleanType }
+        ]
+    }
+    {
+      Name = "direct_table"
+      Columns =
+        [
+          { Name = "str_col"; Type = StringType }
+          { Name = "int_col"; Type = IntegerType }
+          { Name = "float_col"; Type = RealType }
+          { Name = "bool_col"; Type = BooleanType }
+        ]
+    }
+  ]
+
+let anonParams =
+  { AnonymizationParams.Default with
+      TableSettings = Map [ "anon_table", { AidColumns = [ "id" ] } ]
   }
 
+let dataProvider = dummyDataProvider testTables
+let queryContext = QueryContext.make anonParams dataProvider
 
-let dataProvider = dummyDataProvider [ testTable ]
-let queryContext = QueryContext.make AnonymizationParams.Default dataProvider
-
-let aidColIndex = Table.findColumn testTable "int_col" |> fst
-
-let analyzeQuery isAnonymizing queryString =
+let analyzeQuery queryString =
   queryString
   |> Parser.parse
   |> Analyzer.analyze queryContext
-  |> QueryValidator.validateQuery isAnonymizing queryContext.AnonymizationParams
+  |> Normalizer.normalize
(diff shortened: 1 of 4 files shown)


src/OpenDiffix.Core.Tests/QueryValidator.Tests.fs
```diff
@@ -2,33 +2,50 @@ module OpenDiffix.Core.QueryValidatorTests
...
-      [
-        { Name = "str_col"; Type = StringType }
-        { Name = "int_col"; Type = IntegerType } // AID column
-        { Name = "float_col"; Type = RealType }
-        { Name = "bool_col"; Type = BooleanType }
-      ]
+let testTables =
+  [
+    {
+      Name = "anon_table"
```
pdobacz, line 8: you might instead switch the AccessLevel between `Trusted` and `Direct`. Maybe that's more concise.

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -309,73 +309,63 @@ let private bucketExpand aidColumnsExpression =
...
+  { selectQuery with GroupBy = groupBy; Having = having }
+
+let private compileQuery anonParams (selectQuery: SelectQuery) =
   let rangeColumns = collectRangeColumns anonParams selectQuery.From
   let aidColumnsExpression = makeAidColumnsExpression rangeColumns
 
-  let isAnonymizing =
-    anonParams.AccessLevel <> Direct
-    && aidColumnsExpression |> Expression.unwrapListExpr |> List.isEmpty |> not
+  let directAccess =
```
pdobacz, line 358: I'm not necessarily a fan of inverting this condition and calling it `directAccess`. That can be true even despite `AccessLevel <> Direct`, so despite being equivalent, might give rise to some confusion.

Also note, that you're still using the `anonymizing` word in `validate/compileAnonymizingQuery`.

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -309,73 +309,63 @@ let private bucketExpand aidColumnsExpression =
...
+    anonParams.AccessLevel = Direct
+    || aidColumnsExpression |> Expression.unwrapListExpr |> List.isEmpty
 
-  QueryValidator.validateQuery isAnonymizing anonParams.AccessLevel selectQuery
+  QueryValidator.validateStandardQuery selectQuery
```
pdobacz, line 362: not about your change, but I'm also finding the term `standard query` confusing - even despite already knowing what that means. Even more so, that the validation involved is about "non-standard" (in layman terms) concepts like `DiffixLowCount` aggregators.

How about we rename this globally to `direct query` (it's probably spread across the repo, but not very frequent, should be manageable)? Then it lines up with `direct access` nicely, and better reflects what is really happening (if we're using Diffix, we're using it directly).

pdobacz (approved):
Feel free to leave the suggested changes up to my follow-up, in the cases where you agree.

src/OpenDiffix.Core.Tests/QueryValidator.Tests.fs
cristianberneanu: I would prefer to control this from the query statement instead of needing to pass additional arguments to the helper test functions.

src/OpenDiffix.Core/Analyzer.fs
cristianberneanu: Inverting the condition name simplified a bit the corresponding checks. 
I don't feel strongly about it so I will leave it up to you to revert the change or maybe use a different name like `isStandardQuery`.

src/OpenDiffix.Core/Analyzer.fs
cristianberneanu: I'm not convinced that fits well with the intent of this function.
Calling this a `direct query` implies that it isn't an `anonymizing query`. This function should execute common SQL validations that apply to both direct and anonymizing queries. Maybe we should have 3 entry points here: standard SQL checks, direct access checks and anonymizing checks? I will leave it up to you to decide.

cristianberneanu:
I merged this as is so I don't further block you from applying your changes.

src/OpenDiffix.Core/Analyzer.fs
pdobacz: yeah, same issue with `standard query`, we run this even if the query isn't standard. I'll propose something

cristianberneanu: In my head, this somewhat makes sense: the standard SQL validations also apply to anonymizing queries.

pdobacz: To me, the confusion is in the gist of the name `standard query`, in some of the context it appears in here. Also in the fact that `validateStandardQuery` doesn't do checks relevant to anonymizing query. Realizing the latter made it pretty straightforward to untangle, let me show my refactor in a minute...

PR #332, opened 2022-03-03
Refactor Analyzer and QueryValidator

pdobacz:
`QueryValidator` is now strictly a submodule of `Analyzer` (should we arrange the files and module names to reflect that?). As such, it now holds the functionality to `validateGeneralizations`. Also I'm proposing the naming scheme which would liberate us from `standard query`, which IMHO makes the entire construction read a little better.

I couldn't resist the temptation to split SQLNoiseLayers module out, but this is an option piece of the change, I'm not sure about this. It just seemed very separate, similar to how `QueryValidator` is separate.

`SQLNoiseLayer` still is responsible for validation, which it performas while collecting the seed materials, but I think there's no point in moving that over to `QueryValidator`.


src/OpenDiffix.Core/SQLNoiseLayers.fs
```diff
@@ -0,0 +1,42 @@
+module OpenDiffix.Core.SQLNoiseLayers
```
cristianberneanu, line 1: Probably better to drop `SQL` from the module name, as we will likely also include functionality for the dynamic component of noise layers here.

cristianberneanu (approved):

cristianberneanu (approved):

PR #333, opened 2022-03-09
Move data anonymization to leaf subqueries.

### LINKED ISSUE #320: Move anonymization to leaf subqueries.
There are too many features we don't support yet, which makes anonymizing in the top query unfeasible.
Leaf subqueries referencing sensitive table should be the anonymizing ones.
That means no joins, no filters, no unsupported bucketization. Everything post-anonymization should be allowed (including ORDER BY, LIMIT, OFFSET, SELECT with grouping, and HAVING in the anonymizing subquery).

cristianberneanu:
First, the global execution context is dropped and replaced with a global query context plus a per-subquery node anonymization context. Afterwards, the limited support for AID values floating is also dropped and data anonymization is moved to the leaf subquery nodes.

Closes #320.

### DIFF
--- a/src/OpenDiffix.CLI/JsonEncodersDecoders.fs
+++ b/src/OpenDiffix.CLI/JsonEncodersDecoders.fs
@@ -21,7 +21,7 @@ module private rec Encoders =
     | String string -> Encode.string string
     | List values -> Encode.list (values |> List.map encodeValue)
 
-  let rec private typeName =
+  let private typeName =
     function
     | BooleanType -> "boolean"
     | IntegerType -> "integer"
--- a/src/OpenDiffix.CLI/JsonEncodersDecoders.fs
+++ b/src/OpenDiffix.CLI/JsonEncodersDecoders.fs
@@ -27,7 +27,6 @@ module private rec Encoders =
     | IntegerType -> "integer"
     | RealType -> "real"
     | StringType -> "text"
-    | ListType itemType -> typeName itemType + "[]"
     | UnknownType _ -> "unknown"
 
   let private encodeType = typeName >> Encode.string
--- a/src/OpenDiffix.Core.Tests/Expression.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/Expression.Tests.fs
@@ -535,8 +535,15 @@ let colRef2 = ColumnReference(2, RealType)
 
 let evaluate expr = Expression.evaluate testRow expr
 
+let aggContext =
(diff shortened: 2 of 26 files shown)


src/OpenDiffix.CLI/JsonEncodersDecoders.fs
```diff
@@ -27,7 +27,6 @@ module private rec Encoders =
     | IntegerType -> "integer"
     | RealType -> "real"
     | StringType -> "text"
-    | ListType itemType -> typeName itemType + "[]"
```
edongashi, line 30: Function `rec` decorator can be removed.

src/OpenDiffix.Core.Tests/Expression.Tests.fs
```diff
@@ -535,8 +535,15 @@ let colRef2 = ColumnReference(2, RealType)
...
+let aggContext =
+  {
+    AnonymizationParams = AnonymizationParams.Default
+    GroupingLabels = [||]
+    Aggregators = [||]
+  }
+
 let evaluateAggregator aggSpec args =
-  evaluateAggregator (ExecutionContext.makeDefault ()) aggSpec args testRows
+  evaluateAggregator (aggContext, Some { BucketSeed = 0UL }) aggSpec args testRows
```
edongashi, line 546: Potentially dumb question, but can we make the `Some { BucketSeed = 0UL }` part of `aggContext`? Any specific reason to pass them around separately?

cristianberneanu: The two components of this context have different scopes: the aggregation context applies to the current aggregation operation and the anonymization context applies to the current bucket. It seems natural to keep them separate. In the case of evaluating aggregators, we could create another context structure to contain both components, but that adds another level of indirection and pollutes the type namespace. I wasn't convinced it is worth doing until now.

cristianberneanu: The "AnonymizationContext" type used here should be replaced with a "BucketContext" type, to make it more clear it applies to each individual bucket. The two are identical for now (they might not be in the future), so I postponed adding another type to reduce type pollution. But maybe that wasn't the best decision...

edongashi: We already have a ton of "context"s around so the simpler the better!

edongashi (approved):

PR #334, opened 2022-03-14
When casting as boolean, produce Null on invalid value

pdobacz:
Required for https://github.com/diffix/desktop/issues/282

I also decided to throw in a `Trim` in there. I noticed that PostgreSQL handles it like that.

### DIFF
--- a/src/OpenDiffix.Core/Expression.fs
+++ b/src/OpenDiffix.Core/Expression.fs
@@ -190,13 +190,12 @@ let rec evaluateScalarFunction fn args =
     | true, r -> Real r
     | false, _ -> Null
   | Cast, [ String s; String "boolean" ] ->
-    match s.ToLower() with
+    match s.Trim().ToLower() with
     | "true"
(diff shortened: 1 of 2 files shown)


src/OpenDiffix.Core/Expression.fs
```diff
@@ -190,13 +190,13 @@ let rec evaluateScalarFunction fn args =
...
-    match s.ToLower() with
+    match s.Trim().ToLower() with
     | "true"
     | "1" -> Boolean true
     | "false"
     | "0" -> Boolean false
-    | "" -> Null
-    | _ -> failwith "Input value is not a valid boolean string."
+    | ""
+    | _ -> Null
```
edongashi, line 199: Can be merged to only `_ -> Null`

pdobacz: heh, I knew you'd suggest that. I kept separate since cases are different, but it's fine merged too.

edongashi (approved):

PR #335, opened 2022-03-16
Make value hashing consistent with `pg_diffix`.

cristianberneanu:


src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -17,7 +17,8 @@ let cloneHashSet (hashSet: HashSet<'T>) = HashSet<'T>(hashSet, hashSet.Comparer)
 let private invalidArgs (values: Value list) =
   failwith $"Invalid arguments for aggregator: {values}"
 
-let private hashAid (aidValue: Value) = Value.hash aidValue
+let private hashAid (aidValue: Value) =
+  aidValue |> Value.toString |> Hash.string
```
pdobacz, line 21: how consistent is consistent we desire here? Off the top of my head, firstly, I think `reference` appends `L` to the integers and floating point formatting rules might be different.

I'm guessing we have no use-case yet for exact match of the hashes between the 2 implementations, so we don't care about those details, but just making sure.

pdobacz: Actually, I'm confused, it appears that we're hashing memory bytes in `pg_diffix`, not string representations, or am I missing something?

pdobacz (dismissed):
Codewise LGTM, but maybe I am missing something, left one question to clarify.

src/OpenDiffix.Core/Aggregator.fs
cristianberneanu: Good catch, we hash integers directly in `pg_diffix`.

PR #338, opened 2022-04-04
Fix bug where count(value) in the query clashes with *-bucket hook

pdobacz:
Without this fix, the finding function in `CommonTypes` would find both `DiffixCount` (i.e. `count(*), count(value)`) and on two finds, it would return a `None` resulting in an error `Cannot find required DiffixCount aggregator`.

This should fix it.

The `hasOnlyAidArgs` is very opportunistic and brittle, is there a better and still easy way?

### DIFF
--- a/src/OpenDiffix.Core/CommonTypes.fs
+++ b/src/OpenDiffix.Core/CommonTypes.fs
@@ -451,6 +451,11 @@ module Plan =
   let explain (plan: Plan) = toString 0 plan
 
 module AggregationContext =
+  let private hasOnlyAidArgs args =
+    match args with
+    | [ ListExpr _ ] -> true
+    | _ -> false
+
   let private findSingleIndex cond arr =
     arr
     |> Array.indexed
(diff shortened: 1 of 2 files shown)


src/OpenDiffix.Core/CommonTypes.fs
```diff
@@ -459,15 +464,15 @@ module AggregationContext =
...
   let lowCountIndex (aggregationContext: AggregationContext) =
-    match findAggregator DiffixLowCount aggregationContext.Aggregators with
+    match findSingleIndex (fun ((fn, _), _) -> fn = DiffixLowCount) aggregationContext.Aggregators with
     | Some index -> index
     | None -> failwith "Cannot find required DiffixLowCount aggregator"
 
   let diffixCountIndex (aggregationContext: AggregationContext) =
-    match findAggregator DiffixCount aggregationContext.Aggregators with
+    match findSingleIndex
+      // We're looking for `count(*)`, so we cannot confuse a we may find `count(value)` with it.
```
edongashi, line 474: Confusing sentence / typo.

edongashi (approved):

edongashi:
> The hasOnlyAidArgs is very opportunistic and brittle, is there a better and still easy way?

I think count star has only one argument (AIDs), whereas count value has 2 (aids, value).

pdobacz:
> > The hasOnlyAidArgs is very opportunistic and brittle, is there a better and still easy way?
> 
> I think count star has only one argument (AIDs), whereas count value has 2 (aids, value). `Aggregator.fs` L141 verifies this.

Yup, so this little function tests for that, but it might be easy for this function's assumptions to get out of sync with reality.

PR #343, opened 2022-05-05
Implement the --strict-check option and validate anon params

### LINKED ISSUE #342: Prevent setting nonsense anon params and add "strict" option
e.g. `--layer-noise-sd -10`

~NOTE: this is still pending the result of the discussion, whether and how to handle the "strictier" requirements on anon params, namely those that provide safe reasonable anonymization, but some portion of the work can be done regardless~

EDIT: handle the "strict" checks of the anon params (safe reasonable anonymization, as defined by Elm spec) with a new option

pdobacz:
Closes #342 

I thought about more elaborate names, but ended up sticking with something simple

### DIFF
--- a/src/OpenDiffix.Core/CommonTypes.fs
+++ b/src/OpenDiffix.Core/CommonTypes.fs
@@ -149,7 +149,7 @@ type AnonymizationParams =
     Salt: byte []
     Suppression: SuppressionParams
     AccessLevel: AccessLevel
-    StrictCheck: bool
+    Strict: bool
 
     // Count params
     OutlierCount: Interval
@@ -162,7 +162,7 @@ type AnonymizationParams =
       Salt = [||]
       Suppression = SuppressionParams.Default
       AccessLevel = PublishTrusted
-      StrictCheck = true
+      Strict = true
       OutlierCount = Interval.Default
       TopCount = Interval.Default
       LayerNoiseSD = 1.0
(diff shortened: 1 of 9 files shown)


edongashi (dismissed):

src/OpenDiffix.Core/CommonTypes.fs
```diff
@@ -375,6 +381,55 @@ module AnonymizationParams =
       | Some tableSettings -> tableSettings.AidColumns |> List.exists (String.equalsI columnName)
       | None -> false
 
+  /// Fails if any of the anon params does not meet the requirements. Set `strict` to `true` to enforce
+  /// checking if the parameters ensure safe minimum level of anonymization, `false` only for basic checks.
+  let validate anonParams =
+    if anonParams.StrictCheck then
```
edongashi, line 387: Maybe have 2 anonymous records with min values instead of 2 branches? The latter can become out of date more easily.

edongashi:
`strict-check` is a little vague. How about `strict-params` or `strict-config`?

pdobacz:
> `strict-check` is a little vague. How about `strict-params` or `strict-config`?

`strict-params` is not bad, maybe also `strict-params-check`? But it's getting long.

OTOH, we can do "vague by design" and just leave it at `strict`. This would encompass any other notion of "strictness", any measure we might design to prevent the user from making a mistake. On first iteration now, that would just happen to mean strict anon params checking.

edongashi:
Good idea. I'm ok with `strict` or `strict-mode`.

src/OpenDiffix.Core/CommonTypes.fs
pdobacz: hm hm, I can try, but this will be more glue code, and what if we introduce a param which must not be greater than x; then the name like `StrictMinValues` will be out of date. Also if we introduce requirements which aren't simple inequalities, etc. I dunno

pdobacz: :+1: I just did the refactored version, but I'll undo it and keep in stash. I think I do prefer the dumb way for now.

pdobacz:
I've thrown in a commit to validate the size of top/outlier intervals, since I think there's consensus on that it is necessary in `--strict` mode

PR #344, opened 2022-05-10
Use a more compatible version of *-bucket labels for seeding

pdobacz:
Related to https://github.com/diffix/pg_diffix/issues/345 & https://github.com/diffix/pg_diffix/pull/348

`pg_diffix` (or any other Diffix implementation, which will deliver the *-bucket using a pure SQL query-response interface), cannot use `*` for the *-bucket label, for non-string columns.

I propose to go down the same path as `pg_diffix` does, and (at current state of things, just for the sake of seed calculation), use `NULL` for non-string columns.


cristianberneanu (approved):
👍

PR #345, opened 2022-05-11
Ensure deterministic noisy count by picking the largest abs noise

pdobacz:
Accompanies https://github.com/diffix/pg_diffix/pull/351.

This adds a little bit of determinism, where the amount of noise applied to counts might differ, depending on which AID instance is going to be picked in case of draws. With this change, the draws will be handled (almost) deterministically.

### DIFF
--- a/src/OpenDiffix.Core/Anonymizer.fs
+++ b/src/OpenDiffix.Core/Anonymizer.fs
@@ -217,27 +217,18 @@ let private countDistinctFlatteningByAid
 let private anonymizedSum (byAidSum: AidCount seq) =
   let flattening =
     byAidSum
-    |> Seq.sortByDescending (fun aggregate -> aggregate.Flattening)
-    |> Seq.groupBy (fun aggregate -> aggregate.Flattening)
-    |> Seq.head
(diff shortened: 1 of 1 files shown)


src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -229,7 +229,12 @@ let private anonymizedSum (byAidSum: AidCount seq) =
 
   let noise =
     byAidSum
-    |> Seq.maxBy (fun aggregate -> aggregate.NoiseSD)
+    |> Seq.sortByDescending (fun aggregate -> aggregate.NoiseSD)
+    |> Seq.groupBy (fun aggregate -> aggregate.NoiseSD)
+    |> Seq.head
+    // For determinism, resolve draws using maximum absolute noise value.
+    |> snd
+    |> Seq.maxBy (fun aggregate -> Math.Abs(aggregate.Noise))
```
cristianberneanu, line 237: Doesn't `Seq.maxBy (fun aggregate -> aggregate.NoiseSD, Math.Abs(aggregate.Noise))` work?

cristianberneanu (approved):

PR #353, opened 2022-05-19
Add DiffixSum

### LINKED ISSUE #350: DiffixSum aggregate

pdobacz:
Closes #350 

I solved the issue of typing of the `sum` result dynamically, but this feels clunky. I'm wondering if there's an elegant way of extracting the type from the expression? We extract the length of aids dynamically, so I'm kinda guessing it's not easy, but still this makes me wonder.

I'm also unsure, whether `sum(arbitrary expression)` can be type stable, i.e. can we always now upfront, what `arbitrary expression`'s type will be, before starting to execute?

### DIFF
--- a/src/OpenDiffix.Core.Tests/Aggregator.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/Aggregator.Tests.fs
@@ -12,13 +12,17 @@ let randomNullableInteger (random: System.Random) =
   // 10% chance to get a NULL
   if random.Next(10) = 0 then Null else Integer(random.Next(100) |> int64)
 
+let randomNullableReal (random: System.Random) =
+  // 10% chance to get a NULL
+  if random.Next(10) = 0 then Null else Real(random.Next(100) |> float)
+
 let buildAidInstancesSequence numAids (random: System.Random) =
   // Infinite sequence of (Value.List [aid1, aid2, ...])
   Seq.initInfinite (fun _ -> List.init numAids (fun _ -> randomNullableInteger random) |> Value.List)
 
-let buildIntegerSequence (random: System.Random) =
+let buildRealSequence (random: System.Random) =
   // Infinite sequence of (Value.Integer int | Null)
-  Seq.initInfinite (fun _ -> randomNullableInteger random)
+  Seq.initInfinite (fun _ -> randomNullableReal random)
 
 /// Builds a list of given length with aggregator transitions.
 /// Each transition contains AID instances as the first argument
@@ -26,7 +30,7 @@ let buildIntegerSequence (random: System.Random) =
 let makeAnonArgs hasValueArg random numAids length =
   (if hasValueArg then
      // Generates a sequence of [ Value.List [aid1, aid2, ...]; Value.Integer int ]
-     (buildAidInstancesSequence numAids random, buildIntegerSequence random)
+     (buildAidInstancesSequence numAids random, buildRealSequence random)
      ||> Seq.map2 (fun aidInstances argValue -> [ aidInstances; argValue ])
    else
      // Generates a sequence of [ Value.List [aid1, aid2, ...] ]
@@ -39,7 +43,7 @@ let makeAnonArgs hasValueArg random numAids length =
 let makeStandardArgs hasValueArg random length =
   (if hasValueArg then
      // Generates a sequence of [ Value.Integer int ]
-     buildIntegerSequence random |> Seq.map (fun argValue -> [ argValue ])
+     buildRealSequence random |> Seq.map (fun argValue -> [ argValue ])
    else
      // Generates a sequence of [ ]
      Seq.initInfinite (fun _ -> []))
@@ -130,6 +134,10 @@ let testStandard distinct hasArg fn =
 let ``Merging DiffixCount`` () =
   DiffixCount |> testAnon NON_DISTINCT WITH_VALUE_ARG
   DiffixCount |> testAnon NON_DISTINCT WITHOUT_VALUE_ARG
+
+[<Fact>]
+let ``Merging DiffixSum`` () =
+  DiffixSum |> testAnon NON_DISTINCT WITH_VALUE_ARG
 
 [<Fact>]
 let ``Merging distinct DiffixCount`` () =
--- a/src/OpenDiffix.Core.Tests/Aggregator.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/Aggregator.Tests.fs
@@ -26,7 +26,7 @@ let buildRealSequence (random: System.Random) =
 
 /// Builds a list of given length with aggregator transitions.
 /// Each transition contains AID instances as the first argument
-/// and an optional random integer as the second argument.
+/// and an optional random integer (as `Real`) as the second argument.
 let makeAnonArgs hasValueArg random numAids length =
   (if hasValueArg then
      // Generates a sequence of [ Value.List [aid1, aid2, ...]; Value.Integer int ]
@@ -53,16 +53,18 @@ let makeStandardArgs hasValueArg random length =
 /// Verifies that merging 2 separately aggregated sequences is equivalent
 /// to a single aggregation of those 2 sequences concatenated.
 let ensureConsistentMerging ctx fn sourceArgs destinationArgs =
-  let sourceAggregator = Aggregator.create fn
+  let DUMMY_ARGS = [ ListExpr [ ColumnReference(0, IntegerType) ]; ColumnReference(1, RealType) ]
+
+  let sourceAggregator = Aggregator.create (fn, DUMMY_ARGS)
   sourceArgs |> List.iter sourceAggregator.Transition
 
-  let destinationAggregator = Aggregator.create fn
+  let destinationAggregator = Aggregator.create (fn, DUMMY_ARGS)
   destinationArgs |> List.iter destinationAggregator.Transition
 
   destinationAggregator.Merge sourceAggregator
   let mergedFinal = destinationAggregator.Final ctx
 
-  let replayedAggregator = Aggregator.create fn
+  let replayedAggregator = Aggregator.create (fn, DUMMY_ARGS)
   (destinationArgs @ sourceArgs) |> List.iter replayedAggregator.Transition
   let replayedFinal = replayedAggregator.Final ctx
 
--- a/src/OpenDiffix.Core.Tests/Anonymizer.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/Anonymizer.Tests.fs
@@ -7,22 +7,27 @@ open CommonTypes
 
 let companies i =
   let names = [ "Alpha"; "Beta"; "Gamma"; "Delta" ]
-  names |> List.item (i % names.Length)
+  names |> List.item (i % names.Length) |> String
+
+let reals i =
+  let reals = [ 0.1; 0.0; 1.0; -0.01 ]
+  reals |> List.item (i % reals.Length) |> Real
 
 let rows =
   [ 1, 5; 2, 4; 3, 2; 4, 1; 5, 5; 6, 4; 7, 3; 8, 6 ]
   |> List.collect (fun (id, count) -> List.replicate count id)
-  |> List.map (fun id -> [| id |> int64 |> Integer; String "value"; companies id |> String |])
+  |> List.map (fun id -> [| id |> int64 |> Integer; String "value"; companies id; reals id |])
   |> List.append [
-       [| Null; String "value"; String "Alpha" |]
-       [| Integer 8L; Null; String "Alpha" |]
-       [| Integer 9L; String "value"; Null |]
+       [| Null; String "value"; String "Alpha"; Real -100.0 |]
+       [| Integer 8L; Null; String "Alpha"; Null |]
+       [| Integer 9L; String "value"; Null; Real 10.0 |]
      ]
 
 let aidColumn = ColumnReference(0, IntegerType)
 let aidColumnList = ListExpr [ aidColumn ]
 let strColumn = ColumnReference(1, StringType)
 let companyColumn = ColumnReference(2, StringType)
+let realColumn = ColumnReference(3, RealType)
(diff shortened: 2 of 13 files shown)


src/OpenDiffix.Core.Tests/Aggregator.Tests.fs
```diff
@@ -39,7 +43,7 @@ let makeAnonArgs hasValueArg random numAids length =
 let makeStandardArgs hasValueArg random length =
   (if hasValueArg then
      // Generates a sequence of [ Value.Integer int ]
-     buildIntegerSequence random |> Seq.map (fun argValue -> [ argValue ])
+     buildRealSequence random |> Seq.map (fun argValue -> [ argValue ])
```
edongashi, line 46: I wouldn't be surprised if the associative laws of merging don't apply if real numbers are at play because of precision losses.

src/OpenDiffix.Core.Tests/Anonymizer.Tests.fs
```diff
@@ -77,6 +83,25 @@ let ``anon count(col)`` () =
   |> evaluateAggregator diffixCount [ aidColumnList; strColumn ]
   |> should equal (Integer 30L)
 
+[<Fact>]
+let ``anon sum(real col)`` () =
+  // - 1 user with Null real is ignored
```
edongashi, line 88: The list markers don't add much value imo.

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -288,6 +288,115 @@ type private DiffixLowCount() =
...
+  let increaseUnaccountedFor valueIncrease i =
+    let absValueIncrease = abs (valueIncrease)
+
+    if valueIncrease > 0.0 then
+      state.Positive.[i].UnaccountedFor <- state.Positive.[i].UnaccountedFor + absValueIncrease
+    else if valueIncrease < 0.0 then
+      state.Negative.[i].UnaccountedFor <- state.Negative.[i].UnaccountedFor + absValueIncrease
+    else
+      state.Positive.[i].UnaccountedFor <- state.Positive.[i].UnaccountedFor + absValueIncrease
+      state.Negative.[i].UnaccountedFor <- state.Negative.[i].UnaccountedFor + absValueIncrease
```
edongashi, line 312: Isn't `absValueIncrease` zero here? Why add it?

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -288,6 +288,115 @@ type private DiffixLowCount() =
...
+  let increaseSumContribution valueIncrease aidValue i =
+    let absValueIncrease = abs (valueIncrease)
+
+    if valueIncrease > 0.0 then
+      increaseContribution absValueIncrease aidValue state.Positive.[i].AidContributions
+    else if valueIncrease < 0.0 then
+      increaseContribution absValueIncrease aidValue state.Negative.[i].AidContributions
+    else
+      increaseContribution absValueIncrease aidValue state.Positive.[i].AidContributions
+      increaseContribution absValueIncrease aidValue state.Negative.[i].AidContributions
```
edongashi, line 323: Again, zero increase

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -288,6 +288,115 @@ type private DiffixLowCount() =
...
+      match aidValue with
+      // No AIDs, add to unaccounted value
+      | Null
+      | Value.List [] -> increaseUnaccountedFor valueIncrease i
+      // List of AIDs, distribute contribution evenly
+      | Value.List aidValues ->
+        let partialIncrease = valueIncrease / (aidValues |> List.length |> float)
+
+        aidValues
+        |> List.iter (fun aidValue -> increaseSumContribution partialIncrease aidValue i)
```
edongashi, line 339: Maybe we should get rid of this partial increase features completely? We are nowhere near a spec to handle it...

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -288,6 +288,115 @@ type private DiffixLowCount() =
...
+        increaseContributions valueIncrease aidInstances
+    | _ -> failwith "Expecting a list as input"
+
+  member this.State = state
+
+  interface IAggregator with
+    member this.Transition args =
+      match args with
+      | Value.List [] :: _ -> invalidArgs args
+      // Note that we're completely ignoring `Null`, contrary to `count(col)` where it contributed 0.
```
edongashi, line 359:
```suggestion
      // Note that we're completely ignoring `Null`, contrary to `count(col)` where it contributes 0.
```

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -39,12 +39,14 @@ let private mapColumnReference rangeColumns tableName columnName =
   ColumnReference(index, column.Type)
 
 let private mapFunctionExpression rangeColumns fn parsedArgs =
+  let mapAids parsedAids =
+    parsedAids |> List.map (mapExpression rangeColumns) |> ListExpr
```
edongashi, line 43: The function seems general enough to be called `mapExpressions`?

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -53,8 +55,15 @@ let private mapFunctionExpression rangeColumns fn parsedArgs =
...
        | parsedExpr -> aggregateArgs, [ mapExpression rangeColumns parsedExpr ]
 
-     let aids = parsedAids |> List.map (mapExpression rangeColumns) |> ListExpr
-     AggregateFunction(DiffixCount, aggregateArgs), aids :: args
+     AggregateFunction(DiffixCount, aggregateArgs), mapAids parsedAids :: args
+   | AggregateFunction (DiffixSum, aggregateArgs), parsedArg :: parsedAids ->
+     let aggregateArgs, args =
+       match parsedArg with
+       | ParserTypes.Distinct parsedExpr ->
+         { aggregateArgs with Distinct = true }, [ mapExpression rangeColumns parsedExpr ]
```
edongashi, line 63: Do we reject `sum(DISTINCT x)`?

src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -3,6 +3,20 @@ module OpenDiffix.Core.Anonymizer
 open System
 open System.Security.Cryptography
 
+[<RequireQualifiedAccess>]
+type AnonymizedResult<'okType> =
```
edongashi, line 7: Types are always `PascalCase`. `type AnonymizedResult<'T>` would be good enough by convention.

src/OpenDiffix.Core/Expression.fs
```diff
@@ -67,6 +67,7 @@ let typeOfAggregate fn args =
   | DiffixCount -> IntegerType
   | DiffixLowCount -> BooleanType
   | Sum -> RealType
+  | DiffixSum -> RealType
```
edongashi, line 70: You can peak the first arg to get its type.

edongashi:
> I'm wondering if there's an elegant way of extracting the type from the expression?

Can't we match the arg type?

```
let typeOfAggregate fn args =
  match fn with
  ...
  | DiffixSum -> args |> List.head |> Expression.typeOf
```

src/OpenDiffix.Core/Aggregator.fs
pdobacz: I simplified this altogether and forgot to push the change before review :facepalm:. I'll fix with the other fixes

src/OpenDiffix.Core/Analyzer.fs
pdobacz: yes, but done in `QueryValidator` along with `sum(something_complex)`

src/OpenDiffix.Core/Aggregator.fs
pdobacz: If there's consensus, sure, but let's do it in a separate PR, since I think this is orthogonal, and the support for `sum` is already here.

src/OpenDiffix.Core.Tests/Anonymizer.Tests.fs
pdobacz: not sure what you mean. The `-` characters themselves, the contents or...?

Anyhow, I tried to follow the convention from this file.

src/OpenDiffix.Core.Tests/Aggregator.Tests.fs
pdobacz: Hm, is there anything we should do about this?

edongashi: If the tests pass then no. In the future if it happens then we can add a small tolerance.

src/OpenDiffix.Core/Expression.fs
pdobacz: ok, but two questions here:
1/ is it type stable? (see PR description)
2/ what next? Does it fit under `AggregateOptions` so I can next parametrize `Aggregators.DiffixSum` with it? Or should it rather be a trait of the union member `AggregateFunction.DiffixSum`?

pdobacz: Hm, probably answer to 1/ is yes, I'm not sure why I would think otherwise. The issue with the count of AID instances must have sent me off the scent somehow.

edongashi: > can we always know upfront, what arbitrary expression's type will be, before starting to execute?

Yes, we have static type support. If the type can't be resolved to `real` or `integer`, the analyzer should fail to compile.

> so I can next parametrize Aggregators.DiffixSum with it?

Can you inject the required data it in the aggregator constructor? Seems like the most straightforward solution.

pdobacz: > Can you inject the required data it in the aggregator constructor? Seems like the most straightforward solution.

ok, lemme try this and come back to you with a specific question (if any). 

Thanks for the clarification, didn't think about the analyzer failure part.

src/OpenDiffix.Core.Tests/Anonymizer.Tests.fs
edongashi: > The - characters themselves, the contents or...?

Yes, the `-` followed by `//` seem redundant.

src/OpenDiffix.Core/Expression.fs
edongashi: You are right, sorry - I thought we had them in scope.

> I'm now working on including the AggregateArgs in Aggregator.create, but it's a bumpy road

Shouldn't be too difficult, I can make a patch if you like.

pdobacz: I'm close, but not quite there yet. Let me give it some more tries, and we can sit down together next week, if I get stuck

pdobacz: OK, I've got this, thank you!

Now, I'm tempted to simplify the initialization of the Aggregator classes leveraging the `aggArgs` we now have readily available (this not being done was what made me think typed aggregator is not possible in the first place). WDYT?

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -303,23 +303,19 @@ type private DiffixSum() =
   let increaseUnaccountedFor valueIncrease i =
     let absValueIncrease = abs (valueIncrease)
 
-    if valueIncrease > 0.0 then
-      state.Positive.[i].UnaccountedFor <- state.Positive.[i].UnaccountedFor + absValueIncrease
-    else if valueIncrease < 0.0 then
-      state.Negative.[i].UnaccountedFor <- state.Negative.[i].UnaccountedFor + absValueIncrease
-    else
+    if valueIncrease >= 0.0 then
       state.Positive.[i].UnaccountedFor <- state.Positive.[i].UnaccountedFor + absValueIncrease
```
edongashi, line 307: Isn't `>=` redundant because only the `>` case has an effect?

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -381,15 +379,16 @@ type private DiffixSum() =
 
         mergeStateLeg state.Positive otherState.Positive
         mergeStateLeg state.Negative otherState.Negative
-        state.IsReal <- state.IsReal || otherState.IsReal
```
edongashi: Maybe add an assertion instead that both states need to have the same type?

edongashi (approved):

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -406,13 +405,15 @@ let isAnonymizing ((fn, _args): AggregatorSpec) =
   | DiffixSum -> true
   | _ -> false
 
-let create (aggSpec: AggregatorSpec) : T =
+let create (aggSpec: AggregatorSpec, aggArgs: AggregatorArgs) : T =
+  let aggType = Expression.typeOfAggregate (fst aggSpec) aggArgs
```
edongashi, line 409: Variable can be moved to relevant match case in L418.

src/OpenDiffix.Core/Aggregator.fs
pdobacz: yeah, for the unaccounted for that's indeed true. I wanted to keep it parallel to the `increaseContribution` form, but let's do `>` / `<`

edongashi (approved):

src/OpenDiffix.Core/Expression.fs
edongashi: > OK, I've got this, thank you!
> 
> Now, I'm tempted to simplify the initialization of the Aggregator classes leveraging the `aggArgs` we now have readily available (this not being done was what made me think typed aggregator is not possible in the first place). WDYT?
> 
> (if at all, in a follow-up PR, of course)

Sorry, I missed this question. By now I have lost context of what we were doing, but I'm OK with simplifying things.

pdobacz: ha, and I've missed your answer. I can make the follow up PR to demonstrate what I'm getting at.

PR #355, opened 2022-05-23
Remove bucket seed from LCF noise calculation

### LINKED ISSUE #352: Drop SQL component of LCF seed
LCF should depend on AID seed only.

pdobacz:
Closes #352, related to https://github.com/diffix/pg_diffix/pull/360


pdobacz:
Ah, This should go into the `elm` branch, not `master`, I'll fix it tmr morning, but I guess the diff is going to be the same

edongashi (approved):
Does this include the count distinct case?

pdobacz:
> Does this include the count distinct case?

I believe so, I checked this specifically, but I will double-check. The issue with `pg_diffix` was that a rather low-level function was called from the count distinct code

PR #357, opened 2022-05-24
Simplify Aggregators.fs

pdobacz:
Two changes following-up from the DiffixSum commit.

1. Remove `partialIncrease` we spoke about
2. Remove the nullability of State in anonymizing aggregators state. It decreases the noise and branching in the aggregators' code a bit, and we don't have the null to worry about. This has been possible, because of the change that #350 brought, which is the availability of argument expressions in the `Aggregators.create` function. I'm not 100% sold on this yet, but I think overall it does makes things simple. WDYT?

### DIFF
--- a/src/OpenDiffix.Core/Aggregator.fs
+++ b/src/OpenDiffix.Core/Aggregator.fs
@@ -124,14 +124,8 @@ type private DiffixCount() =
 
       match aidValue with
       // No AIDs, add to unaccounted value
-      | Null
-      | Value.List [] -> aidState.UnaccountedFor <- aidState.UnaccountedFor + valueIncrease
-      // List of AIDs, distribute contribution evenly
-      | Value.List aidValues ->
-        let partialIncrease = (float valueIncrease) / (aidValues |> List.length |> float)
-
-        aidValues
-        |> List.iter (fun aidValue -> increaseContribution partialIncrease aidValue aidContributions)
+      | Null -> aidState.UnaccountedFor <- aidState.UnaccountedFor + valueIncrease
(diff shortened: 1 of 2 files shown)


src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -124,14 +124,8 @@ type private DiffixCount() =
...
-      | Null
-      | Value.List [] -> aidState.UnaccountedFor <- aidState.UnaccountedFor + valueIncrease
-      // List of AIDs, distribute contribution evenly
-      | Value.List aidValues ->
-        let partialIncrease = (float valueIncrease) / (aidValues |> List.length |> float)
-
-        aidValues
-        |> List.iter (fun aidValue -> increaseContribution partialIncrease aidValue aidContributions)
+      | Null -> aidState.UnaccountedFor <- aidState.UnaccountedFor + valueIncrease
+      | Value.List _ -> failwith "Lists of AIDs and distributing of contributions are not supported."
```
cristianberneanu, line 124: Why is this union case kept if it is not supported anymore?

cristianberneanu (approved):

src/OpenDiffix.Core/Aggregator.fs
pdobacz: You mean to remove `Value.List` altogether on the type system level?

cristianberneanu: Yes.

edongashi: No, please keep it. The upcoming histogram needs it.

pdobacz: okay, so merging the PR as is. Thx!

PR #358, opened 2022-05-26
Add DiffixCountNoise, unoptimized, unanonymized

pdobacz:
As title, this doesn't try to leverage the coexistence of `count(x)` and `count_noise(x)` in one query, just does the job twice.

I did some versions of handling the switch and the typing, this seems to work good enough, I'm not 100% sure though.

### DIFF
--- a/src/OpenDiffix.Core/Aggregator.fs
+++ b/src/OpenDiffix.Core/Aggregator.fs
@@ -50,6 +50,10 @@ let private increaseContribution valueIncrease aidValue (aidMap: Dictionary<AidH
 
   aidMap.[aidHash] <- updatedContribution
 
+type private RequestedOutput =
+  | Value
+  | Noise
+
 // ----------------------------------------------------------------
 // Aggregators
 // ----------------------------------------------------------------
@@ -107,7 +111,7 @@ type private Sum() =
 
     member this.Final(_aggContext, _anonContext) = state
 
-type private DiffixCount(aidsCount) =
+type private DiffixCount(aidsCount, requestedOutput) =
   let mutable state: Anonymizer.AidCountState array =
     Array.init aidsCount (fun _ -> { AidContributions = Dictionary<AidHash, float>(); UnaccountedFor = 0.0 })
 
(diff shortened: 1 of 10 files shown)


src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -107,7 +111,7 @@ type private Sum() =
 
     member this.Final(_aggContext, _anonContext) = state
 
-type private DiffixCount(aidsCount) =
+type private DiffixCount(aidsCount, requestedOutput) =
```
cristianberneanu, line 114: The language already provides a standard mechanism for partially overriding an object's behavior: inheritance. Isn't that sufficient? I think it would be more elegant than a constructor flag.

src/OpenDiffix.Core/QueryValidator.fs
```diff
@@ -30,21 +30,34 @@ let private validateAllowedAggregates query =
...
     | FunctionExpr (AggregateFunction (_otherAggregate, _), _) ->
-      failwith "Only count and sum aggregates are supported in anonymizing queries."
+      failwith "Only count, count_noise and sum aggregates are supported in anonymizing queries."
     | _ -> ()
   )
 
 let private allowedCountUsage query =
+  let message =
+    "Only count(*), count(column), count_noise(*), count_noise(column)"
+    + " and count(distinct column) are supported in anonymizing queries."
```
cristianberneanu, line 43: This type of error message is not scalable: the more aggregators we support, the longer it gets. And we also need to update it each time. Something more generic would be better here, like: "Invoked aggregator is not supported in anonymizing queries.".

src/OpenDiffix.Core/QueryValidator.fs
```diff
@@ -30,21 +30,34 @@ let private validateAllowedAggregates query =
...
   query
   |> visitAggregates (
     function
-    | FunctionExpr (AggregateFunction (Count, _), args) ->
-      match args with
-      | []
-      | [ ColumnReference _ ] -> ()
-      | _ -> failwith "Only count(*), count(column) and count(distinct column) are supported in anonymizing queries."
+    | FunctionExpr (AggregateFunction (Count, opts), args) -> checkArgs args
+    | FunctionExpr (AggregateFunction (CountNoise, opts), args) ->
```
cristianberneanu, line 55: Why not explicitly match for `Distinct` in a separate `match` branch?
`Count` and `CountNoise` can also be matched at the same time.

src/OpenDiffix.Core/Aggregator.fs
pdobacz: Inheritance is exactly what I was trying to avoid, until we get a sense of the abstractions required here (e.g. for the optimized way of handling the computation). I feel like it would be pretty likely that those abstractions would need to be redesigned, but we can do via inheritance as well if you prefer.

src/OpenDiffix.Core/QueryValidator.fs
pdobacz: yeah, was thinking the same and wondering is it already the time to cut this. Current form has the advantage of spelling out what is supported, but that won't last.

src/OpenDiffix.Core/Aggregator.fs
cristianberneanu: I don't think I understand what you mean. In my mind, the abstraction already exists (`IAggregator`) and I suggest we keep using it as it is.

cristianberneanu: To be clear: I suggest that `CountNoise` inherits from `Count` and only overrides the `Final` function.

PR #360, opened 2022-05-26
Add DiffixAvg

### LINKED ISSUE #359: Add DiffixAvg
rewrite to `diffix_sum/diffix_count`

pdobacz:
Closes #359.

I've also noticed some problems with the anonymized `sum` and fixed here (mainly that sum wasn't rewritten to `DiffixSum` in anonymized queries, also a minor typing glitch)

### DIFF
--- a/src/OpenDiffix.Core/Analyzer.fs
+++ b/src/OpenDiffix.Core/Analyzer.fs
@@ -42,11 +42,19 @@ let private mapFunctionExpression rangeColumns fn parsedArgs =
   let mapAids parsedAids =
     parsedAids |> List.map (mapExpression rangeColumns) |> ListExpr
 
+  let mapAvg aggregateArgs parsedArgs sumFunction countFunction =
+    let sum = mapFunctionExpression rangeColumns (AggregateFunction(sumFunction, aggregateArgs)) parsedArgs
+    let count = mapFunctionExpression rangeColumns (AggregateFunction(countFunction, aggregateArgs)) parsedArgs
+    let castSum = FunctionExpr(ScalarFunction ScalarFunction.Cast, [ sum; Constant(String "real") ])
+    ScalarFunction ScalarFunction.Divide, [ castSum; count ]
+
   (match fn, parsedArgs with
    | AggregateFunction (Count, aggregateArgs), [ ParserTypes.Star ] -> //
      AggregateFunction(Count, aggregateArgs), []
    | AggregateFunction (CountNoise, aggregateArgs), [ ParserTypes.Star ] -> //
      AggregateFunction(CountNoise, aggregateArgs), []
+   | AggregateFunction (Avg, aggregateArgs), parsedArgs -> //
+     mapAvg aggregateArgs parsedArgs Sum Count
    | AggregateFunction (DiffixLowCount, aggregateArgs), parsedAids ->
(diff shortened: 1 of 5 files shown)


src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -42,11 +42,19 @@ let private mapFunctionExpression rangeColumns fn parsedArgs =
...
+    let castSum = FunctionExpr(ScalarFunction ScalarFunction.Cast, [ sum; Constant(String "real") ])
+    ScalarFunction ScalarFunction.Divide, [ castSum; count ]
+
   (match fn, parsedArgs with
    | AggregateFunction (Count, aggregateArgs), [ ParserTypes.Star ] -> //
      AggregateFunction(Count, aggregateArgs), []
    | AggregateFunction (CountNoise, aggregateArgs), [ ParserTypes.Star ] -> //
      AggregateFunction(CountNoise, aggregateArgs), []
+   | AggregateFunction (Avg, aggregateArgs), parsedArgs -> //
+     mapAvg aggregateArgs parsedArgs Sum Count
```
edongashi, line 57: Why do we need this one? Is this the direct access version of avg?

src/OpenDiffix.Core/Analyzer.fs
```diff
@@ -42,11 +42,19 @@ let private mapFunctionExpression rangeColumns fn parsedArgs =
   let mapAids parsedAids =
     parsedAids |> List.map (mapExpression rangeColumns) |> ListExpr
 
+  let mapAvg aggregateArgs parsedArgs sumFunction countFunction =
+    let sum = mapFunctionExpression rangeColumns (AggregateFunction(sumFunction, aggregateArgs)) parsedArgs
+    let count = mapFunctionExpression rangeColumns (AggregateFunction(countFunction, aggregateArgs)) parsedArgs
+    let castSum = FunctionExpr(ScalarFunction ScalarFunction.Cast, [ sum; Constant(String "real") ])
```
edongashi, line 48: Is the extra cast dropped later if it's redundant?

src/OpenDiffix.Core/Analyzer.fs
pdobacz: I complicated the story. This line is executed for `SELECT avg(age) FROM...` regardless of direct/anonymized access. The other added line here with `DiffixAvg` is executed for `SELECT diffix_avg(age)` in direct access.

src/OpenDiffix.Core/Analyzer.fs
pdobacz: Correction, it does gets dropped, I'm blind. For some reason it doesn't for `cast(1.0, 'real')`, I'll investigate further.

But I guess we're good here, in the context of `avg`?

edongashi: > For some reason it doesn't for cast(1.0, 'real')

Should be folded by the compiler/normalizer. Worth fixing as it's no-op.

edongashi: > But I guess we're good here, in the context of avg?

Yes.

> For some reason it doesn't for cast(1.0, 'real')

Interesting, worth investigating and fixing.

pdobacz: Ok, I've got it, will file a follow up PR to fix

PR #361, opened 2022-05-26
Add support for simple `WHERE` clauses.

### LINKED ISSUE #326: Support simple filters.
We can support simple filters of the form `bucket_expression = value` by producing the same bucket seeds that an equivalent grouping query would produce. The query
```SQL
SELECT count(*) FROM table WHERE round_by(age, 10) = 20
```
should result in a anonymized count with the same SQL seed as the query
```
SELECT round_by(age, 10), count(*) FROM table GROUP BY 1
```
has for the bucket with the label `20`.

cristianberneanu:
Closes #326.

### DIFF
--- a/src/OpenDiffix.Core/CommonTypes.fs
+++ b/src/OpenDiffix.Core/CommonTypes.fs
@@ -205,7 +205,7 @@ type Plan =
 // Executor
 // ----------------------------------------------------------------
 
-type AnonymizationContext = { BucketSeed: Hash }
+type AnonymizationContext = { BucketSeed: Hash; BaseLabels: Value list }
 
 type AggregationContext =
   {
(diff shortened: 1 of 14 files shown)


src/OpenDiffix.Core/CommonTypes.fs
```diff
@@ -205,7 +205,7 @@ type Plan =
 // Executor
 // ----------------------------------------------------------------
 
-type AnonymizationContext = { BucketSeed: Hash }
+type AnonymizationContext = { BucketSeed: Hash; BaseLabels: Value list }
```
edongashi, line 208: Are BaseLabels supposed to be implicit values of grouping labels inferred by WHERE `x=const`?

edongashi (approved):

src/OpenDiffix.Core/CommonTypes.fs
cristianberneanu: Yes. They are labels that implicitly apply to the entire anonymized result set.

PR #364, opened 2022-05-27
Add simpler count distinct AID variant to anonymizer

edongashi:
I want to avoid the complex count distinct version in favor of a simple and specialized `count(distinct aid)` aggregator. Because I find it hard to understand the flattening code, I need feedback on this particular function before moving on with the histogram aggregate.

### DIFF
--- a/src/OpenDiffix.Core/Anonymizer.fs
+++ b/src/OpenDiffix.Core/Anonymizer.fs
@@ -263,26 +263,28 @@ let private anonymizedSum (byAidSum: AidCount seq) =
 // Public API
 // ----------------------------------------------------------------
 
+/// Returns whether a single AID value set is low count.
+let aidIsLowCount (anonParams: AnonymizationParams) numAids aidSeed =
+  if numAids < anonParams.Suppression.LowThreshold then
+    true
+  else
+    let thresholdNoise =
+      [ aidSeed ]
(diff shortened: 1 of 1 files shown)


src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -326,6 +326,24 @@ let countDistinct
...
+  (aidSeed: Hash)
+  =
+  let noise =
+    [ anonContext.BucketSeed; aidSeed ]
+    |> generateNoise anonParams.Salt "noise" anonParams.LayerNoiseSD
+
+  let noisyCount = (float numAids + noise) |> Math.roundAwayFromZero |> int64
+
+  if noisyCount < anonParams.Suppression.LowThreshold then
+    AnonymizedResult.NotEnoughAIDVs
```
cristianberneanu, line 343: `noisyCount` should be clamped to `LowThreshold` in this case. The `NotEnoughAIDVs` return value should be produced only when there aren't enough AID values for flattening, which we skip in this simplified version of the function.

cristianberneanu (approved):

cristianberneanu:
> But in order to determine the histogram bin we need to know how many rows an AID has contributed. With multiple AIDs I'm not sure how to do the counting...

The same way we do `count(distinct column)`, but replace `column` with the counted AID.

PR #365, opened 2022-05-30
Make sure normalization of cast(1.01 as real) works

pdobacz:
This line https://github.com/diffix/reference/blob/master/src/OpenDiffix.Core/Normalizer.fs#L19 evaluates the `cast(1.01 as real)` which used to be invalid cast.

### DIFF
--- a/src/OpenDiffix.Core/Expression.fs
+++ b/src/OpenDiffix.Core/Expression.fs
@@ -204,6 +204,10 @@ let rec evaluateScalarFunction fn args =
   | Cast, [ Real r; String "text" ] -> r.ToString(doubleStyle) |> String
   | Cast, [ Boolean b; String "text" ] -> b.ToString().ToLower() |> String
 
+  | Cast, [ Integer i; String "integer" ] -> Integer i
+  | Cast, [ Real r; String "real" ] -> Real r
+  | Cast, [ Boolean b; String "boolean" ] -> Boolean b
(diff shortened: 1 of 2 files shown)


src/OpenDiffix.Core/Expression.fs
```diff
@@ -204,6 +204,10 @@ let rec evaluateScalarFunction fn args =
   | Cast, [ Real r; String "text" ] -> r.ToString(doubleStyle) |> String
   | Cast, [ Boolean b; String "text" ] -> b.ToString().ToLower() |> String
 
+  | Cast, [ Integer i; String "integer" ] -> Integer i
```
edongashi, line 207: I would add this to line 199 alongside numeric casts.

edongashi (approved):

edongashi:
I don't think this is right. This needs to be handled in the normalizer.

edongashi (approved):

src/OpenDiffix.Core/Expression.fs
pdobacz: I'm not sure what pattern/ordering of clauses you are after? I don't recognize any pattern in the preexisting clauses, so the new clauses were added at the end and clustered together (as they all fall into the "noop cast" category).

edongashi: Ok with the above pattern it looks good. I thought they were grouped by source type.

PR #366, opened 2022-05-30
sum_noise(col) and count_noise(distinct col)

### LINKED ISSUE #362: Support `sum_noise` and `count_noise(distinct ...)`
... and `avg_noise`?

pdobacz:
Closes #362, without `avg_noise` still pending discussion.

This is turning a little bit copy-pastive, but once we see whether we want to support `avg_noise` or not, we can think about some refactorings maybe.

I've spent some time testing it manually, but I didn't include any tests which would cover those exact new paths. I think we have enough coverage regardless, so no need to multiply this test code. Let me know, if you think otherwise.

### DIFF
--- a/src/OpenDiffix.Core/Anonymizer.fs
+++ b/src/OpenDiffix.Core/Anonymizer.fs
@@ -10,7 +10,7 @@ type AnonymizedResult<'T> =
 
 type AidCountState = { AidContributions: Dictionary<AidHash, float>; mutable UnaccountedFor: float }
 
-type CountResult = { AnonymizedSum: int64; NoiseSD: float }
+type CountResult<'T> = { AnonymizedSum: 'T; NoiseSD: float }
 
 type SumState =
   {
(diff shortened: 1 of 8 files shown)


src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -356,16 +358,21 @@ let sum (anonParams: AnonymizationParams) (anonContext: AnonymizationContext) (p
...
+        | [||] -> (0.0, 0.0)
+        | nonEmpty -> anonymizedSum nonEmpty
 
     // Using `anonymizedSum` separately for positive and negative, we ensure that we pick the appropriate
     // amount of flattening and noise for each leg, and only later combine the results.
-    let positive = anonymizedSumOnNonEmpty byAidPositive
-    let negative = anonymizedSumOnNonEmpty byAidNegative
+    let (positive, positiveNoiseSD) = anonymizedSumOnNonEmpty byAidPositive
+    let (negative, negativeNoiseSD) = anonymizedSumOnNonEmpty byAidNegative
+    let noiseSD = Math.Sqrt(positiveNoiseSD ** 2.0 + negativeNoiseSD ** 2.0)
```
pdobacz, line 368: this is a formula for _independent_ variables, and currently my best shot, but I need to give it some more thought.

`positive` and `negative` aren't independent (the noisy `AnonymizedSums`, that is), so I guess their noise isn't either.

Still TBD, I'll make a self-note to kick off a discussion, unless someone has a quick argument I'm missing here.

cristianberneanu: This makes sense, but shouldn't it also contain a check for the `NotEnoughAIDVs` case?

cristianberneanu (approved):

src/OpenDiffix.Core/Anonymizer.fs
cristianberneanu: I now saw the previous check that requires both (positive and negative) partial results to be valid. This is not correct: it is enough to have one valid partial result in order to return a final valid result.

pdobacz: it does, it's several lines above. If neither positive nor negative batch have enough AIDVs for a particular AID instance, it will return `AnonymizedResult.NotEnoughAIDVs`. In `Aggregator.fs` this will get translated to `Null`.

At this line here, SDs of both batches are the maximal ones where there were enough AIDVs.

pdobacz: > I now saw the previous check that requires both (positive and negative) partial results to be valid. This is not correct: it is enough to have one valid partial result in order to return a final valid result.

No, it fails if both positive and negative are invalid. If there are no negative values whatsoever (typical `age` column), it will succeed, if all AID instances have enough AIDVs. Also the `negative` will be 0 and its noise will be 0.0.

PR #372, opened 2022-06-08
Add noiseless count_histogram aggregate

edongashi:
Future todos:

1. Add noise.
2. Require bin size to be a money value in untrusted mode.
3. Require the argument to be an AID
4. ~~Handle NULL values?~~
5. Move to `Anonymizer` module?

### DIFF
--- a/src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
@@ -446,13 +446,13 @@ type Tests(db: DBFixture) =
       "SELECT sum(z.age) FROM (SELECT t.age FROM customers JOIN customers AS t ON true) z WHERE z.age=0"
 
   [<Fact>]
-  let ``Require AID argument for count_distinct`` () =
+  let ``Require AID argument for count_histogram`` () =
     analyzeTrustedQuery "SELECT count_histogram(id) FROM customers" |> ignore
     analyzeTrustedQuery "SELECT count_histogram(id, 3) FROM customers" |> ignore
     assertTrustedQueryFails "SELECT count_histogram(age, 3) from customers" "count_histogram requires an AID argument."
 
   [<Fact>]
-  let ``Require constant integer for count_distinct bin size`` () =
+  let ``Require constant integer for count_histogram bin size`` () =
     let errorMsg = "count_histogram bin size must be a constant positive integer."
     assertTrustedQueryFails "SELECT count_histogram(age, -3) from customers" errorMsg
     assertTrustedQueryFails "SELECT count_histogram(age, age) from customers" errorMsg
--- a/src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
@@ -446,6 +446,19 @@ type Tests(db: DBFixture) =
       "SELECT sum(z.age) FROM (SELECT t.age FROM customers JOIN customers AS t ON true) z WHERE z.age=0"
 
   [<Fact>]
+  let ``Require AID argument for count_distinct`` () =
+    analyzeTrustedQuery "SELECT count_histogram(id) FROM customers" |> ignore
+    analyzeTrustedQuery "SELECT count_histogram(id, 3) FROM customers" |> ignore
+    assertTrustedQueryFails "SELECT count_histogram(age, 3) from customers" "count_histogram requires an AID argument."
+
+  [<Fact>]
+  let ``Require constant integer for count_distinct bin size`` () =
+    let errorMsg = "count_histogram bin size must be a constant positive integer."
+    assertTrustedQueryFails "SELECT count_histogram(age, -3) from customers" errorMsg
+    assertTrustedQueryFails "SELECT count_histogram(age, age) from customers" errorMsg
+    assertTrustedQueryFails "SELECT count_histogram(age, 'text') from customers" errorMsg
+
+  [<Fact>]
   let ``Detect queries with disallowed generalizations in untrusted access level`` () =
     assertUntrustedQueryFails
       "SELECT substring(city, 2, 2) from customers"
@@ -471,6 +484,14 @@ type Tests(db: DBFixture) =
       "SELECT width_bucket(age, 2, 200, 5) from customers"
       "Generalization used in the query is not allowed in untrusted access level."
 
+    assertUntrustedQueryFails
+      "SELECT width_bucket(age, 2, 200, 5) from customers"
+      "Generalization used in the query is not allowed in untrusted access level."
+
+    assertUntrustedQueryFails
+      "SELECT count_histogram(id, 3) from customers"
+      "count_histogram bin size must be a money-aligned value (1, 2, 5, 10, ...)."
+
   [<Fact>]
   let ``Analyze queries with allowed generalizations in untrusted access level`` () =
     analyzeUntrustedQuery "SELECT substring(city, 1, 2) from customers" |> ignore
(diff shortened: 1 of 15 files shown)


src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -382,6 +382,125 @@ type private DiffixSumNoise(summandType, aidsCount) =
...
+
+  member this.State = state
+
+
+type private CountHistogramAidTracker = { mutable RowCount: int64; LowCount: DiffixLowCount }
+
+type private CountHistogramBin = { mutable AidCount: int64; LowCount: DiffixLowCount }
+
+type private DiffixCountHistogram(aidsCount, binSize: int64 option) =
+  let state = Dictionary<AidHash, CountHistogramAidTracker>()
```
edongashi, line 431: We can have an unhashed `Dictionary<Value, CountHistogramAidTracker>` here but this would lose consistency with a future postgres implementation.

src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
```diff
@@ -187,11 +187,70 @@ type Tests(db: DBFixture) =
   [<Fact>]
   let ``query 14 - avg_noise parity`` () =
     let expected = runQuery "SELECT sum_noise(age) / count(age) as avg_noise FROM customers_small GROUP BY city"
-
     let queryResult = runQuery "SELECT avg_noise(age) FROM customers_small GROUP BY city"
 
     queryResult |> should equal expected
 
+  let ``query 15 - count histogram`` () =
```
pdobacz, line 194: this is my copy pasta, but maybe you could fix this - I've put in `query 14` twice above, so yours is 16

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -382,6 +382,125 @@ type private DiffixSumNoise(summandType, aidsCount) =
...
+    | None
+    | Some 1L -> count
+    | Some binSize -> (float count / float binSize) |> floor |> int64 |> (*) binSize
+
+  interface IAggregator with
+    member this.Transition args =
+      match args with
+      | Null :: _ -> () // Ignore NULLs
+      | aid :: _ -> state |> Dictionary.increment aid
+      | _ -> invalidArgs args
```
pdobacz, line 400: Do I understand correctly, that multi-AID isn't supported or is it working transparently as a list of AID instances? If the former, we could put in a specific error case for this, if the latter maybe we should spell this `aids` or `aidList` here, to make it clear we support multi-AID.

(same comment for the Diffix version, with the additional twist, that we need to differentiate from the `aidInstances` there. So perhaps, a specific name to put here and there would be `countedAid` / `countedAids`?)

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -382,6 +382,125 @@ type private DiffixSumNoise(summandType, aidsCount) =
...
+
+type private CountHistogramBin = { mutable AidCount: int64; LowCount: DiffixLowCount }
+
+type private DiffixCountHistogram(aidsCount, binSize: int64 option) =
+  let state = Dictionary<AidHash, CountHistogramAidTracker>()
+
+  let makeAidTracker () =
+    { RowCount = 0; LowCount = DiffixLowCount(aidsCount) }
+
+  let generalize count =
```
pdobacz, line 436: same as for the non-Diffix one?

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -410,6 +530,15 @@ let create (aggSpec: AggregatorSpec, aggArgs: AggregatorArgs) : T =
     else
       0
 
+  let unpackConstInt index =
```
pdobacz, line 533: I got stuck on this for a while, somehow thinking there's much more to it than it actually is. Given the error message is specific anyway, maybe better to call this function specifically like `unpackGeneralizationWidthArgAt`.

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -382,6 +382,125 @@ type private DiffixSumNoise(summandType, aidsCount) =
...
+
+      let suppressBin = { AidCount = 0L; LowCount = DiffixLowCount(aidsCount) }
+
+      let highCountBins =
+        bins
+        |> Seq.choose (fun pair ->
+          let binLabel = pair.Key
+          let bin = pair.Value
+
+          if Anonymizer.isLowCount aggContext.AnonymizationParams bin.LowCount.State then
```
pdobacz, line 485: do you think moving this bit to `Anonymizer.countHistogram` would be justified or overengineered. I feel like for consistency, some portion of this code should already be living there, even if it's going to be rather trivial at the moment.

pdobacz (approved):
LGTM, just a bunch of minor suggestions. Very impressive work, the count_histogram felt always a bit intimidating to me, it's great to see this come to life :clap:.

src/OpenDiffix.Core/Aggregator.fs
edongashi: > Do I understand correctly, that multi-AID isn't supported or is it working transparently as a list of AID instances?

We pass the list of AIDs like usual, but only _one_ AID can be counted. The rest are used for LCF during pass 2 of processing. As a reminder:

- Pass 1: Gather a `map<aid, (rows_contributed, sibling_aid_sets[])>` during agg transitions where `sibling_aids` are AID sets from rows associated with given `aid`.
- Pass 2: Invert the map to `map<num_rows, (num_aids_with_such_rows, aid_instances[])>`, filter bins with low count `aid_instances`.
- Pass 3: Finalize the map to a sorted list of 2-tuples `(count, num_aids)`.

> So perhaps, a specific name to put here and there would be countedAid / countedAids?

Good idea!

src/OpenDiffix.Core/Aggregator.fs
edongashi: Yes, I'm slightly annoyed to pollute top level module but let's lift up.

src/OpenDiffix.Core/Aggregator.fs
pdobacz: gotcha, so it's going to be `countedAid` (singular) or any better alternative you might find, correct?

pdobacz: I mean, we can't abuse this by passing a _single_ AID which has a type `List`?

edongashi: > I mean, we can't abuse this by passing a _single_ AID which has a type `List`?

Right now the anon version accepts the following args: `(aidInstances: List, countedAid: Aid, binSize?: integer)`.

`countedAid` is guaranteed to be a member of `aidInstances` but it's simpler to just pass it again.

src/OpenDiffix.Core/Aggregator.fs
edongashi: No thoughts on this. I would prefer to leave for later until we add the additional missing features.

edongashi: I moved only the noise adding part in `Anonymizer` the Dictionary handling remains there because I don't see how it can be reused.

src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
```diff
@@ -445,6 +445,19 @@ type Tests(db: DBFixture) =
     analyzeDirectQuery
       "SELECT sum(z.age) FROM (SELECT t.age FROM customers JOIN customers AS t ON true) z WHERE z.age=0"
 
+  [<Fact>]
+  let ``Require AID argument for count_distinct`` () =
```
pdobacz, line 449: should be `for count_histogram` x2

pdobacz (approved):
one typo and lgtm

src/OpenDiffix.Core.Tests/Analyzer.Tests.fs
edongashi: :man_facepalming: thanks!

pdobacz (approved):

PR #378, opened 2022-07-13
Bug fixes for count histogram

### LINKED ISSUE #376: Finalizer of AnonCountHistogram is not idempotent
It just occurred to me that we should not tamper with trans state in finalizer.

edongashi:
Fixes #376 and 2 noise generation bugs.


pdobacz (approved):

edongashi:
:grimacing: let me know if it works now

PR #379, opened 2022-07-14
Move single quoting to be done in List stringifying only

pdobacz:
In order for 25 and "25" to seed the same. The single quotes in List are necessary to avoid ambiguities in lists like `["a", "b", "c"]` and `["a,b","c"]`.


pdobacz:
CI fails, because I needed to update formatter locally :/ .

src/OpenDiffix.Core/CommonTypes.fs
```diff
@@ -300,8 +300,8 @@ module Expression =
     | Boolean b -> b.ToString()
     | Integer i -> i.ToString()
     | Real r -> r.ToString()
-    | String s -> String.quoteSingle s
-    | List values -> "[" + (values |> List.map valueToString |> String.joinWithComma) + "]"
+    | String s -> s
+    | List values -> "[" + (values |> List.map (valueToString >> String.quoteSingle) |> String.joinWithComma) + "]"
```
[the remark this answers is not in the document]
pdobacz: gotcha, thx!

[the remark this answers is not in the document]
pdobacz: I force pushed this right away, that change was definitely one step too far

src/OpenDiffix.Core/Value.fs
```diff
@@ -16,8 +16,8 @@ let rec toString value =
   | Boolean false -> "f"
   | Integer i -> i.ToString()
   | Real r -> r.ToString()
-  | String s -> String.quoteSingle s
-  | List values -> "[" + (values |> List.map toString |> String.joinWithComma) + "]"
+  | String s -> s
+  | List values -> "[" + (values |> List.map (toString >> String.quoteSingle) |> String.joinWithComma) + "]"
```
edongashi, line 20: We need to pass context data or call another function, otherwise this will quote everything, including numbers and other lists. A 2D list of numbers would print as `[ '[1,2]', '[3,4]' ]` which is not desired.

edongashi: OTOH this could work:

```fs
let rec toString value =
  match value with
  | Null -> "NULL"
  | Boolean true -> "t"
  | Boolean false -> "f"
  | Integer i -> i.ToString()
  | Real r -> r.ToString()
  | String s -> String.quoteSingle s
  | List values -> "[" + (values |> List.map toString |> String.joinWithComma) + "]"
  
  and toStringListItem value =
    match value with
    | String s -> String.quoteSingle s
    | value -> toString value
  ```

pdobacz: yup, this makes sense, I think this check's all the boxes

edongashi (approved):

edongashi:
I just noticed something. I think `double.ToString()` is culture sensitive, resulting in a string say `1,500` in a German thread culture. This will mess up our list representation because of the ambiguous separator.

PR #380, opened 2022-08-10
Simplify anonymizer.

cristianberneanu:
Some refactoring in preparation of implementing flattened data recovery.

### DIFF
--- a/src/OpenDiffix.Core/Aggregator.fs
+++ b/src/OpenDiffix.Core/Aggregator.fs
@@ -112,7 +112,7 @@ type private Sum() =
     member this.Final(_aggContext, _anonContext) = state
 
 type private DiffixCount(aidsCount) =
-  let mutable state: Anonymizer.AidCountState array =
+  let mutable state: Anonymizer.ContributionsState array =
     Array.init aidsCount (fun _ -> { AidContributions = Dictionary<AidHash, float>(); UnaccountedFor = 0.0 })
 
   /// Increases contribution of all AID instances.
@@ -171,7 +171,7 @@ type private DiffixCount(aidsCount) =
         else
           int64 aggContext.AnonymizationParams.Suppression.LowThreshold
 
-      match Anonymizer.sum aggContext.AnonymizationParams anonContext state with
+      match Anonymizer.count aggContext.AnonymizationParams anonContext state with
       | None -> Integer minCount
       | Some { AnonymizedSum = value } -> value |> Math.roundAwayFromZero |> int64 |> max minCount |> Integer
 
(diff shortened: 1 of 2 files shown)


src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -246,8 +246,8 @@ type private DiffixCountDistinctNoise(aidsCount) =
       let anonContext = unwrapAnonContext anonContext
 
       match Anonymizer.countDistinct aggContext.AnonymizationParams anonContext this.AidsCount this.State with
-      | Anonymizer.AnonymizedResult.NotEnoughAIDVs -> Null
-      | Anonymizer.AnonymizedResult.Ok { NoiseSD = noiseSD } -> Real noiseSD
+      | None -> Null
```
pdobacz, line 249: IIRC this has been given a name on purpose, to have the reasoning behind the `Null` case spelled out. Otherwise it's just too many `None` and `Null` cases, making the code hard to follow

pdobacz (commented):
2 items to discuss

src/OpenDiffix.Core/Aggregator.fs
cristianberneanu: I didn't find it helpful at all. It acts like a renamed `Option` type, but with a lot more verbosity.
It provides a single error case, for which we don't do anything special, except returning `Null`.
I think it hinders more than it helps.

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -362,11 +362,21 @@ type private DiffixSum(summandType, aidsCount) =
...
+        else
+          Math.roundAwayFromZero >> int64 >> Integer
+
+      let positiveResult = Anonymizer.sum aggContext.AnonymizationParams anonContext state.Positive
+      let negativeResult = Anonymizer.sum aggContext.AnonymizationParams anonContext state.Negative
 
-      match Anonymizer.sum aggContext.AnonymizationParams anonContext state isReal with
-      | Anonymizer.AnonymizedResult.NotEnoughAIDVs -> Null
-      | Anonymizer.AnonymizedResult.Ok { AnonymizedSum = value } -> value
+      match positiveResult, negativeResult with
```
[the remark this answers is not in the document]
cristianberneanu: Good point! I missed this semantic during the refactoring, I will revert back to using separate handlers for `sum` and `count`.

src/OpenDiffix.Core/Aggregator.fs
pdobacz: > It acts like a renamed `Option` type, but with a lot more verbosity

That's exactly what it is intended to be and I found it very helpful, in contrast to the obscure `Null`s and `None`s being passed around here. It spared me a lot of head-scratching and look-ups.

Note that our course of action in these `match` clauses is very specific to our logic, so a specific value is better than a generic (`None`). Same goes for the conditions in `Anonymizer` which emit these `NotEnoughAIDVs`.

pdobacz (approved):

src/OpenDiffix.Core/Aggregator.fs
cristianberneanu: I don't understand what we get by explicitly exposing the reason for not being able to produce an anonymized result to the callers. It is irrelevant to them, since there is only one possible cause and they also don't take any specific action on it. I find it adds noise to the codebase for little to no gain.

PR #381, opened 2022-08-12
Enable star bucket and LED hooks in CLI app.

### LINKED ISSUE #336: Support *-bucket and LED hooks in CLI
You should be able to use https://github.com/diffix/desktop/blob/master/anonymizer/OpenDiffix.Service/Program.fs#L179-L181 and neighboring code as a template.

Optionally provide a toggle to switch *-bucket on and off and an option to customize the `*` character.

**NOTE** using the *-bucket hook requires `count(*)` to be present in the query. Think about the defaults and documentation in this context

**NOTE 2** global aggregation is also slightly problematic `SELECT count(*) FROM customers  ;` fails with `Cannot find required DiffixLowCount aggregator`, because `lcf` is not injected.

cristianberneanu:
The `StarBucket.hook` callback is updated in order to support adding the star bucket to 
the output buckets. Hooks are not run anymore over the global bucket.

Closes #336.

### DIFF
--- a/src/OpenDiffix.Core/Executor.fs
+++ b/src/OpenDiffix.Core/Executor.fs
@@ -55,7 +55,10 @@ let private invokeHooks aggregationContext anonymizationContext hooks buckets =
   match anonymizationContext with
   | None -> buckets
   | Some anonymizationContext ->
-    List.fold (fun buckets hook -> hook aggregationContext anonymizationContext buckets) buckets hooks
+    if aggregationContext.GroupingLabels.Length > 0 then
+      List.fold (fun buckets hook -> hook aggregationContext anonymizationContext buckets) buckets hooks
+    else
+      buckets // don't run hooks for global bucket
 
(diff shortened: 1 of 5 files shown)


src/OpenDiffix.Core/Executor.fs
```diff
@@ -55,7 +55,10 @@ let private invokeHooks aggregationContext anonymizationContext hooks buckets =
   match anonymizationContext with
   | None -> buckets
   | Some anonymizationContext ->
-    List.fold (fun buckets hook -> hook aggregationContext anonymizationContext buckets) buckets hooks
+    if aggregationContext.GroupingLabels.Length > 0 then
```
edongashi, line 58: The hooks do these checks themselves (at least LED does).

edongashi (approved):

src/OpenDiffix.Core/Executor.fs
cristianberneanu: Both hooks assert that the `low-count` aggregator is present, which it won't be for global aggregation. This is needed to avoid a crash for such queries.

PR #383, opened 2022-08-16
Fix parsing of Floats like 1.101 and 1.lots-of-digits

pdobacz:
I picked it up when using `WHERE x < 40.63` for debugging - it was parsing as `46.3`.

I've made an initial implementation fixing the one already there with some more code, but it started to become a considerable chunk to handle all the cases. So I redid the search and found the `numberLiteral` parser in `FParsec` - I hope this works.

To be sure we're covered, I added some extra test cases. Possibly repetitive since we're using `FParsec` now, but I think worth to have in case we stop doing so.

### DIFF
--- a/src/OpenDiffix.Core/Parser.fs
+++ b/src/OpenDiffix.Core/Parser.fs
@@ -51,19 +51,18 @@ module QueryParser =
 
   let alias = word "AS" >>. identifier
 
-  // This custom numbers parser is needed as both pint64 and pfloat are eager
-  // to the point of not being possible to combine. pint64 would parse 1.2 as 1,
-  // and pfloat would parse 1 as 1.0.
+  let numberFormat =
+    NumberLiteralOptions.AllowMinusSign
+    ||| NumberLiteralOptions.AllowFraction
+    ||| NumberLiteralOptions.AllowExponent
+
   let number =
-    pint64 .>>. opt (pchar '.' >>. many (pchar '0') .>>. opt pint32) .>> spaces
-    |>> fun (wholeValue, decimalPartOption) ->
-          match decimalPartOption with
-          | None -> Expression.Integer wholeValue
-          | Some (leadingZeros, Some decimalValue) ->
-            let divisor = List.length leadingZeros + 1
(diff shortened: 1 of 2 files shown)


src/OpenDiffix.Core/Parser.fs
```diff
@@ -51,19 +51,18 @@ module QueryParser =
...
-            let divisor = List.length leadingZeros + 1
-            let decimalPart = (float decimalValue) / (float <| pown 10 divisor)
-            Expression.Float(float wholeValue + decimalPart)
-          | Some (_leadingZeros, None) -> Expression.Float(float wholeValue)
+    numberLiteral numberFormat "number" .>> spaces
+    |>> fun nl ->
+          if nl.IsInteger then
+            Expression.Integer(int64 nl.String)
+          else
+            Expression.Float(float nl.String)
```
edongashi, line 65: Does `float nl.String` perform culture sensitive parsing?

edongashi (approved):

src/OpenDiffix.Core/Parser.fs
pdobacz: according to some SO answer - no, but I couldn't find a more reliable source

pdobacz: but, I've found `FParsec` to use `System.Double.Parse(nl.String, System.Globalization.CultureInfo.InvariantCulture)` instead, in its own `pfloat`, so I guess that would be more consistent

pdobacz: I decided to force push just this little detail

edongashi: I might have already asked this question. I think we override thread culture to force it to the invariant one.

edongashi: ^ I checked and that's the case. BUT, we do that in `main`, therefore a custom CLI client such as desktop can miss that particular detail and result in weird behavior.

edongashi (changes requested):
Let's revert to `float x` since that's what we do elsewhere as well.

edongashi (approved):

PR #384, opened 2022-08-18
Fix avg/avg_noise when anonymized count is 0

pdobacz:
Sibling of https://github.com/diffix/pg_diffix/pull/437.

The `nullif` has been added _without_ support for `Value.List`, which works in PostgreSQL. I started doing so, but realized, that validating the correct types (which PostgreSQL does) for lists is going to be pretty big and not worth it.

### DIFF
--- a/src/OpenDiffix.Core/Expression.fs
+++ b/src/OpenDiffix.Core/Expression.fs
@@ -54,7 +54,13 @@ let typeOfScalarFunction fn args =
       | Constant (String "real") -> RealType
       | Constant (String "boolean") -> BooleanType
       | _ -> failwith "Unsupported cast destination type name."
-  | NullIf -> args |> List.head |> typeOf
+  | NullIf ->
+    args
+    |> List.map typeOf
+    |> function
+      | [ IntegerType; RealType ] -> RealType
+      | [ RealType; IntegerType ] -> RealType
+      | argsTypes -> List.head argsTypes
 
 /// Resolves the type of a set function expression.
 let typeOfSetFunction fn _args =
@@ -216,11 +222,8 @@ let rec evaluateScalarFunction fn args =
   | Cast, [ Real r; String "text" ] -> r.ToString(doubleStyle) |> String
   | Cast, [ Boolean b; String "text" ] -> b.ToString().ToLower() |> String
 
-  | NullIf, [ Boolean x; Boolean y ] -> if x = y then Null else Boolean x
-  | NullIf, [ Integer x; Integer y ] -> if x = y then Null else Integer x
-  | NullIf, [ Real x; Real y ] -> if x = y then Null else Real x
-  | NullIf, [ String x; String y ] -> if x = y then Null else String x
-  | NullIf, [ Null; Null ] -> Null
+  | NullIf, [ x; y ] when x = y -> Null
+  | NullIf, [ x; y ] -> x
 
   | _ -> failwith $"Invalid usage of scalar function '%A{fn}'."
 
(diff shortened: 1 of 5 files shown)


src/OpenDiffix.Core/Expression.fs
```diff
@@ -215,6 +216,12 @@ let rec evaluateScalarFunction fn args =
   | Cast, [ Real r; String "text" ] -> r.ToString(doubleStyle) |> String
   | Cast, [ Boolean b; String "text" ] -> b.ToString().ToLower() |> String
 
+  | NullIf, [ Boolean x; Boolean y ] -> if x = y then Null else Boolean x
+  | NullIf, [ Integer x; Integer y ] -> if x = y then Null else Integer x
+  | NullIf, [ Real x; Real y ] -> if x = y then Null else Real x
+  | NullIf, [ String x; String y ] -> if x = y then Null else String x
+  | NullIf, [ Null; Null ] -> Null
```
edongashi, line 223: Can't we just generalize to 2 cases:

```fs
NullIf, [ x, y ] when x = y -> Null
NullIf, [ x, _y ] -> x
```

Maybe even the Exliir way works: `NullIf, [ x; x ] -> Null` (not sure if F# supports it)

The F# `Value` ADT will have an (by value) equality operator out of the box so no need to worry about deep comparison.

edongashi (dismissed):

src/OpenDiffix.Core/Expression.fs
pdobacz: Heh, Elixir way doesn't work, tried that by muscle memory first.

But the 2-cases way doesn't break on disparate types, which PostgreSQL does break on. If we don't need the type check and have `nullif(1, 'a')` be just `1`, we can do that. 

Hm, on second thought, PostgreSQL has no problem with `nullif(1.1, 1)`, doing a cast somewhere there. However, in our case, the shallow `=` of the 2 case solution will still not do that trick, we would need to have a bunch of special cases (for example, PostgreSQL's `nullif(1.0, 1)` produces NULL). Hard for me to judge.

edongashi: We can have a general `Value.equivalent` or `Value.softEquals` function which handles the real and integer comparison. But in any case, for `count == 0` we have clearly defined types and we don't care about SQL/Postgres compliance.

edongashi: I say strict equality is fine here considering your current code already does that.

pdobacz: kk, I'll simplify then

pdobacz: lol, I've been scratching my head for a while, why would our simplified `nullif(Integer 1, Real 1.0)` resolve to Null. It turns out there's a generic coercion for binary functions matching before the `NullIf` case:

```
// Treat mixed integer/real binary operations as real/real operations.
| _, [ Integer i; Real r ] -> evaluateScalarFunction fn [ Real(double i); Real r ]
| _, [ Real r; Integer i ] -> evaluateScalarFunction fn [ Real r; Real(double i) ]
```

I guess we can keep this behavior, this is in line with PostgreSQL and common sense.

edongashi: Hah, I haven't been aware of this either!

> I guess we can keep this behavior, this is in line with PostgreSQL and common sense.

Yes, looks good.

edongashi (approved):

PR #387, opened 2022-08-24
Minor tweak to simplify String sorting and align to PostgreSQL

pdobacz:
Some tests revealed that the previous tweaks to the string sorting rules went too far. So I'm taking the opportunity to simplify and align with PG some more.


pdobacz:
ah, hold on a second, there's one more detail, when we're at it

pdobacz:
> ah, hold on a second, there's one more detail, when we're at it

OK, sorry, what I wanted to try doesn't work. So that's it, PR ready for review.

cristianberneanu (approved):

PR #390, opened 2022-09-06
Recover and redistribute flattened outliers data.

cristianberneanu:
Data dropped during flattening is recovered and proportionally redistributed over the anonymized buckets, in order to minimize the total distortion per column.

### DIFF
--- a/src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
@@ -23,6 +23,7 @@ type Tests(db: DBFixture) =
       OutlierCount = { Lower = 1; Upper = 1 }
       TopCount = { Lower = 1; Upper = 1 }
       LayerNoiseSD = 0.
+      RecoverOutliers = true
     }
 
   let noisyAnonParams = { AnonymizationParams.Default with TableSettings = tableSettings }
--- a/src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
+++ b/src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
@@ -289,17 +289,13 @@ type Tests(db: DBFixture) =
       "SELECT count(*) FROM customers_small GROUP BY city ORDER BY 1"
 
   [<Fact>]
-  let ``Filtering and grouping doesn't change results`` () =
+  let ``Equivalent filtering and grouping doesn't change seed`` () =
     equivalentQueries
-      "SELECT count(*) FROM customers GROUP BY city HAVING city = 'Berlin'"
+      "SELECT count(*) FROM customers WHERE city = 'Berlin' GROUP BY city"
       "SELECT count(*) FROM customers WHERE city = 'Berlin'"
 
     equivalentQueries
-      "SELECT count(*) FROM customers GROUP BY city, round_by(age, 10) HAVING city = 'Berlin' AND round_by(age, 10) = 20"
-      "SELECT count(*) FROM customers WHERE city = 'Berlin' AND round_by(age, 10) = 20"
-
-    equivalentQueries
-      "SELECT count(*) FROM customers WHERE round_by(age, 10) = 20 GROUP BY city HAVING city = 'Berlin'"
+      "SELECT count(*) FROM customers WHERE city = 'Berlin' AND round_by(age, 10) = 20 GROUP BY city, round_by(age, 10)"
       "SELECT count(*) FROM customers WHERE city = 'Berlin' AND round_by(age, 10) = 20"
 
   [<Fact>]
--- a/src/OpenDiffix.Core/Aggregator.fs
+++ b/src/OpenDiffix.Core/Aggregator.fs
@@ -87,7 +87,7 @@ type private Count() =
     member this.Merge aggregator =
       state <- state + (castAggregator<Count> aggregator).State
 
-    member this.Final(_aggContext, _anonContext, _outliersAggregrator) = Integer state
+    member this.Final(_aggContext, _anonContext, _outliersAggregator) = Integer state
 
 type private CountDistinct() =
   let state = HashSet<Value>()
@@ -104,7 +104,7 @@ type private CountDistinct() =
     member this.Merge aggregator =
       state.UnionWith((castAggregator<CountDistinct> aggregator).State)
 
-    member this.Final(_aggContext, _anonContext, _outliersAggregrator) = state.Count |> int64 |> Integer
+    member this.Final(_aggContext, _anonContext, _outliersAggregator) = state.Count |> int64 |> Integer
 
 type private Sum() =
   let mutable state = Null
@@ -124,7 +124,7 @@ type private Sum() =
     member this.Merge aggregator =
       (this :> IAggregator).Transition [ (castAggregator<Sum> aggregator).State ]
 
-    member this.Final(_aggContext, _anonContext, _outliersAggregrator) = state
+    member this.Final(_aggContext, _anonContext, _outliersAggregator) = state
 
 type private DiffixCount(aidsCount) =
   let mutable state: Anonymizer.ContributionsState array =
@@ -177,9 +177,9 @@ type private DiffixCount(aidsCount) =
         )
       )
 
-    member this.Final(aggContext, anonContext, outliersAggregrator) =
+    member this.Final(aggContext, anonContext, outliersAggregator) =
       let anonContext = unwrapAnonContext anonContext
-      let getState = fun aggregrator -> (castAggregator<DiffixCount> aggregrator).State
+      let getState = fun aggregator -> (castAggregator<DiffixCount> aggregator).State
 
       let minCount =
         if Array.isEmpty aggContext.GroupingLabels && List.isEmpty anonContext.BaseLabels then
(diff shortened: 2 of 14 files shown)


src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -362,8 +381,10 @@ type private DiffixSum(summandType, aidsCount) =
       mergeContributions state.Positive srcState.Positive
       mergeContributions state.Negative srcState.Negative
 
-    member this.Final(aggContext, anonContext) =
+    member this.Final(aggContext, anonContext, outliersAggregrator) =
       let anonContext = unwrapAnonContext anonContext
+      let getPositiveState = fun aggregrator -> (castAggregator<DiffixSum> aggregrator).State.Positive
+      let getNegativeState = fun aggregrator -> (castAggregator<DiffixSum> aggregrator).State.Negative
```
edongashi, line 387: Typo in `aggregrator`

edongashi: Also saw it in a few other places with CTRL-F.

src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -151,12 +171,20 @@ let inline private aidFlattening
       [ anonContext.BucketSeed; aidContributions |> Seq.map fst |> seedFromAidSet ]
       |> generateNoise anonParams.Salt "noise" noiseSD
 
+    let outliers =
```
pdobacz, line 174: Two comments:
1. How about we call this `outlierFlatteningByAid` or something more verbose across the board?
2. Can we move this bit up and substitute `outlierReplacement` and `flattening` for summing of this array? I have a feeling like we're calculating the same thing twice here

pdobacz (commented):
I've got preliminary comments after first scan, but I need to take a more detailed pass, as it's quite big and complex.

Aside of the inline comments, a question: how do we switch this off?

src/OpenDiffix.Core/Aggregator.fs
cristianberneanu: Thanks, seems that my brain miss-fired so hard when it came to writing this word that it completely replaced the correct spelling in my head... 🙄

src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
```diff
@@ -289,17 +289,13 @@ type Tests(db: DBFixture) =
       "SELECT count(*) FROM customers_small GROUP BY city ORDER BY 1"
 
   [<Fact>]
-  let ``Filtering and grouping doesn't change results`` () =
+  let ``Equivalent filtering and grouping doesn't change seed`` () =
     equivalentQueries
-      "SELECT count(*) FROM customers GROUP BY city HAVING city = 'Berlin'"
+      "SELECT count(*) FROM customers WHERE city = 'Berlin' GROUP BY city"
```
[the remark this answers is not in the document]
cristianberneanu: It is necessary, because when data is filtered out before anonymization there are no outliers to redistribute, resulting in a different output. The same effect happens with LED, so the tested behavior wasn't correct for some time.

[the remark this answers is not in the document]
pdobacz: > when data is filtered out before anonymization there are no outliers to redistribute

just for my understanding - I thought there are outliers, but there aren't other buckets where to distribute them to. Is this what you've meant or am I missing sth?

Other than that, the new form of the assertion seems "easy" to satisfy, so to speak. As in:  the GROUP BY can be ignored by the system under test and we're testing equivalence of identical queries. I'm not sure I'm making a point here.

(I guess we're testing whether the user can't get two samples out of a `GROUP BY` query and a `WHERE` query, correct? BTW Is it possible to break the code in a way that this assertion fails?)

pdobacz:
> > how do we switch this off?
> 
> Didn't implement an off switch yet, as it doesn't seem to make the results worse.

An off switch would allow us to easily keep validating this. Also the ability to keep redistribution off could be useful when implementing and testing features orthogonal to it - like the datetime support - especially until redistribution is implemented on `pg_diffix` end.

These are weak arguments, but maybe worth thinking the off switch through as a thought experiment? Is there a way to bypass a single very last step (comment something out), to switch off? I'll try to find this out on the second pass.

cristianberneanu:
> Also the ability to keep redistribution off could be useful when implementing and testing features orthogonal to it - like the datetime support - especially until redistribution is implemented on pg_diffix end.

Fair enough, I will add toggle to switch it off.

src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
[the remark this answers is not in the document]
pdobacz: OK, this makes sense. OTOH doesn't this (i.e. the offset between the `WHERE` and no-`WHERE` queries) provide a way to reconstruct the flattening?

pdobacz (commented):
2 more comments, submitting before EOD so GH doesn't drop them. I'll continue tmr morning. Apologies for going in small batches, but there is a lot to digest here.

src/OpenDiffix.Core.Tests/QueryEngine.Tests.fs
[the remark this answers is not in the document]
cristianberneanu: The recovered outliers are themselves anonymized before redistribution. 
One can reconstruct the noisy total value of the outliers, but that information is safe to report.

src/OpenDiffix.Core/Anonymizer.fs
```diff
@@ -127,19 +147,23 @@ let inline private aidFlattening
...
     let topGroupSum =
       sortedAidContributions
       |> Seq.skip outlierCount
       |> Seq.take topCount
       |> Seq.sumBy snd
 
     let topGroupAverage = (float topGroupSum) / (float topCount)
-    let outlierReplacement = topGroupAverage * (float outlierCount)
+
+    let droppedContributions =
```
pdobacz, line 158: this is pretty nitpicky, but we're not dropping contributions but altering them - understanding the word "contribution" as a whole "thing" that a given AID contributes. I have no elegant alternative though... `flatteningByAID` or `aidFlattenings` to match other terms aroung this file?

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -50,9 +50,24 @@ let private increaseContribution valueIncrease aidValue (aidMap: Dictionary<AidH
 
   aidMap.[aidHash] <- updatedContribution
 
-type private RequestedOutput =
-  | Value
-  | Noise
+let private mergeOutliers outliers (contributions: Anonymizer.ContributionsState array) =
+  outliers
+  |> Array.iteri (fun aidIndex aidOutliers ->
```
pdobacz, line 55: After the change here I see that in `aidIndex` in this line and `aid` below, AID means a different things. It would be easier to read this if we were explicit here and indicate `aidInstanceIndex` is meant.

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -50,9 +50,24 @@ let private increaseContribution valueIncrease aidValue (aidMap: Dictionary<AidH
...
+    let aidContributions = contributions.[aidIndex].AidContributions
+
+    aidOutliers
+    |> Array.iter (fun (aid, contribution) ->
```
pdobacz, line 59: in light of the other comment from `Anonymizer.fs`, this could be `(aid, flattening)` - if I am following the variables correctly

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -50,9 +50,24 @@ let private increaseContribution valueIncrease aidValue (aidMap: Dictionary<AidH
 
   aidMap.[aidHash] <- updatedContribution
 
-type private RequestedOutput =
-  | Value
-  | Noise
+let private mergeOutliers outliers (contributions: Anonymizer.ContributionsState array) =
```
pdobacz, line 53: I admit I'm finding this part hard to follow - following what is what, and where do we take the data from and which data we mutate... The problem is in the combination of this function here with the `>>` and `getState` construct used in the `Final` members - I have to jump around the file (and `Anonymizer.fs` to follow the data around.

I have however no alternative off-hand. I wish there was a way to code it so it reads more top-to-bottom like a recipe - "take this out of that and put into this other thing".

Maybe an idea would be to duplicate `mergeOutliers` as members of the Aggregator classes first, make them contain all the logic and glue code involved and only then try to abstract out common stuff?

pdobacz (commented):
OK, I think I got how this works. I have some naming suggestions to consider / discuss. Only part I can't fully grok is the `mergeOutliers` part - I have only a vague idea, what it's supposed to do.

src/OpenDiffix.Core/Anonymizer.fs
cristianberneanu: These are the altered contributions that are not included in the aggregated result. They are removed from the total, so calling them "dropped" seems logical to me.

src/OpenDiffix.Core/Aggregator.fs
cristianberneanu: This is the left-over contribution, I think "droppedContribution" makes more sense.

src/OpenDiffix.Core/Aggregator.fs
cristianberneanu: Sorry the code seems obtuse, I struggled with it for a while and couldn't find a more elegant version.
This part is common to count and sum aggregators, I don't see how it would help if we duplicate it.
`getState` has to be called twice each time, so we still need to separate it from this part.

src/OpenDiffix.Core/Aggregator.fs
```diff
@@ -172,19 +188,22 @@ type private DiffixCount(aidsCount) =
           int64 aggContext.AnonymizationParams.Suppression.LowThreshold
 
       match Anonymizer.count aggContext.AnonymizationParams anonContext state with
-      | None -> Integer minCount
-      | Some { AnonymizedSum = value } -> value |> Math.roundAwayFromZero |> int64 |> max minCount |> Integer
+      | None ->
+        Option.iter (getState >> mergeOutliers (contributionsToArray state)) outliersAggregator
```
pdobacz, line 192: late thought - shouldn't we subtract `minCount` here? I'm guessing this means we're treating all contributions as outliers, but since we're reporting the `minCount`, the flattening should not include this figure.

cristianberneanu: Good point! Reporting a positive value for the count on insufficient entities will result in an upwards distortion of the total after redistribution. But the outliers are also flattened separately before redistribution, which will distort the total downwards. The two effects should cancel each other somewhat. And with the code being too complicated already, I'm reluctant to do anything about it.

src/OpenDiffix.Core/Aggregator.fs
pdobacz: I can't see the improvement I suggested yesterday anymore, you're right.

pdobacz (approved):

PR #391, opened 2022-09-08
`date_trunc` and simple `DateTime` support

pdobacz:
Sibling of https://github.com/diffix/pg_diffix/pull/447. 

Contrary to `pg_diffix`, this seems to have shipped with the bonus of `WHERE date_trunc = ...` working out of the box.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -88,3 +88,7 @@ resharper_fsharp_multi_line_lambda_closing_newline = true
 fsharp_multi_line_lambda_closing_newline = true
 resharper_fsharp_strict_mode = false
 fsharp_strict_mode = false
+
+[data/generate.fsx]
+resharper_fsharp_max_array_or_list_width = 120
+fsharp_max_array_or_list_width = 120
--- a/data/generate.fsx
+++ b/data/generate.fsx
@@ -14,6 +14,10 @@ type Field =
   | Text of string
   | Integer of int64
   | Real of float
+  | Timestamp of DateTime
+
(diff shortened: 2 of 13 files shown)


.editorconfig
```diff
@@ -88,3 +88,7 @@ resharper_fsharp_multi_line_lambda_closing_newline = true
 fsharp_multi_line_lambda_closing_newline = true
 resharper_fsharp_strict_mode = false
 fsharp_strict_mode = false
+
+[data/generate.fsx]
```
pdobacz, line 92: This keeps the formatting of the hardcoded table templates sane without too much modification to that file

data/generate.fsx
```diff
@@ -14,23 +14,28 @@ type Field =
...
+  "\"" + string.Replace("\"", "\"\"") + "\""
 
 let fieldToString field =
   match field with
   | Null -> "NULL"
   | Text value -> $"'{value}'"
   | Integer value -> string value
   | Real value -> string value
+  | Timestamp value ->
+    $"""'{value.ToString("yyyy-MM-ddTHH:mm:ss", System.Globalization.CultureInfo.InvariantCulture)}'"""
```
cristianberneanu, line 29: Maybe this should also be a function?

cristianberneanu (approved):

data/generate.fsx
pdobacz: I decided to offload this work to the `Value.Timestamp` stringification.

cristianberneanu (approved):

PR #397, opened 2023-03-16
Support the new 'Adaptive Buckets' anonymization method.

cristianberneanu:


src/OpenDiffix.Core/Utils.fs
```diff
@@ -86,3 +86,17 @@ module Hash =
...
   let roundAwayFromZero (x: float) : float =
     System.Math.Round(x, MidpointRounding.AwayFromZero)
+
+let TIMESTAMP_REFERENCE = DateTime(1800, 1, 1, 0, 0, 0, DateTimeKind.Utc)
+
+// Encodes `s` as a double-precision floating point number, obtained by hashing.
+// Uses only a subset of representable floating point to avoid large/small numbers.
+// For columns with many unique values collisions are possible, but collisions numerous enough
+// to cross a reasonable LCF threshold rather unlikely.
+let hashStringToFloat (s: String) =
```
edongashi, line 96: We recently changed it to map to a value `0..1`, but either will be replaced soon.

src/OpenDiffix.Core/Planner.fs
```diff
@@ -149,4 +121,37 @@ let private filterJunk targetList plan =
...
+  let orderByExpressions = query.OrderBy |> List.map (fun (OrderBy(expression, _, _)) -> expression)
+  let expressions = query.Having :: selectedExpressions @ orderByExpressions
+  let aggregators = expressions |> collectAggregates |> List.distinct
+  let groupingLabels = query.GroupBy |> List.distinct
+
+  let expressionProjector, rowsAggregatorPlanner =
+    if groupingLabels.IsEmpty && aggregators.IsEmpty then
+      match query.AnonymizationContext with
+      | Some anonymizationContext when anonymizationContext.AnonymizationParams.UseAdaptiveBuckets ->
+        projectExpression selectedExpressions, planAdaptiveBuckets selectedExpressions anonymizationContext
```
edongashi, line 134: Do we verify what `selectedExpressions` can contain? Is it the same restrictions that apply to group labels?

What if we have an expr such as `2 * floor_by(age, 10)`? A smart planner might pull the `floor_by` into AB plan and `2 *` to a parent projection. This is rather complex so we might want to rely on clients doing further projections in an upper query.

edongashi (approved):

src/OpenDiffix.Core.Tests/AdaptiveBuckets/Common.Tests.fs
```diff
@@ -0,0 +1,43 @@
+module OpenDiffix.Core.AdaptiveBuckets.CommonTests
```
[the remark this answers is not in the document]
cristianberneanu: What do you mean? This tests module is still present in the main branch.

src/OpenDiffix.Core/Planner.fs
cristianberneanu: Restrictions are validated by the analyzer (and yes, they are the same as for the grouping expressions).
An expression like `2 * floor_by(age, 10)` wouldn't be allowed. Or are you referring to something else?

src/OpenDiffix.Core.Tests/AdaptiveBuckets/Common.Tests.fs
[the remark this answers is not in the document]
cristianberneanu: Ok, but I will update the main version first, so that we can just replace the files afterwards, once we reach a stable codebase.

